PR #8, opened 2015-04-22
added support for (conditional) group overrides. 

marcelveldt:
Added support for group overrides which will solve the issues that exist after the changes in Isengard that a modal dialog should first be closed before opening any other windows.

As discussed here: http://forum.kodi.tv/showthread.php?tid=224683

Usage:

Add an extra entry in your skin's overrides.xml file:

<groupoverride group="powermenu" condition="Window.IsActive(DialogButtonMenu.xml)">Close</groupoverride>

group is the name of the menugroup, for example mainmenu or powermenu etc.

### DIFF
--- a/resources/lib/datafunctions.py
+++ b/resources/lib/datafunctions.py
@@ -199,6 +199,17 @@ class DataFunctions():
             # Get the action
             action = node.find( "action" )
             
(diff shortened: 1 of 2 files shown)


BigNoid:
Besides fixing that issue, could be useful for other usecases as well, thx!

PR #9, opened 2015-04-24
fixed small issue with hardcoded string used instead of language string

marcelveldt:
http://forum.kodi.tv/showthread.php?tid=178294&pid=1972303#pid1972303


marcelveldt:
also the fix for the nonetype error (nullpointer) in case skins do use skinshortcuts but aren't using the overrides file.

 as discussed here:
https://github.com/BigNoid/script.skinshortcuts/issues/10

marcelveldt:
Once merged, could you submit the update to the kodi repo please ?

BigNoid:
Looks good, thx. Will submit a pr after testing.

PR #11, opened 2015-05-23
Added support for <propertypattern>

bambi73:
Added support for custom shortcut properties patterns. 
Feature is described in Advanced Usage.txt point 14.

### DIFF
--- a/resources/lib/xmlfunctions.py
+++ b/resources/lib/xmlfunctions.py
@@ -1,5 +1,5 @@
 # coding=utf-8
-import os, sys, datetime, unicodedata
+import os, sys, datetime, unicodedata, re
 import xbmc, xbmcgui, xbmcvfs, xbmcaddon
 import xml.etree.ElementTree as xmltree
 from xml.sax.saxutils import escape as escapeXML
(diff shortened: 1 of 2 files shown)


resources/lib/xmlfunctions.py
```diff
@@ -717,13 +717,61 @@ def buildElement( self, item, Tree, groupName, visibilityCondition, profileVisib
...
+        propertyPatterns = {}
+        
+        if overrides is not None:
+            propertyPatternElements = overrides.getroot().findall("propertypattern")
+            for propertyPatternElement in propertyPatternElements:
+                propertyName = propertyPatternElement.get("property")
+                propertyGroup = propertyPatternElement.get("group")
+              
+                if propertyName and propertyGroup and propertyPatternElement.text:
+                    if propertyGroup == group:
```
phil65: would be nice you would do "if not ...: continue" instead of super-deep nesting :)

phil65:
looks fine apart from cosmetics (I prefer to use continue instead of deep nesting of conditionals)

bambi73:
Reversed conditions and added continue to remove some nesting

phil65:
thx.

PR #13, opened 2015-06-20
Templates

ghost:
Adds ability to generate gui xml elements based on templates. See templates.txt in commit for full details.

Also a fix for long-standing bug whereby only additional properties for the currently logged in profile were loaded when building menu, and were applied to menu items for all profiles. (Can do as a separate PR if you'd prefer.)


BigNoid:
The idea of templates is very nice, thx!
Will give this some test runs.
@phil65 mind taking a look at the code?

phil65:
not needed. He´s the original author and knows what he is doing :)

BigNoid:
Okay :)

PR #16, opened 2015-06-21
add list property to the output xml to be used for filling a dynamic list

marcelveldt:
This was a little request by @HitcherUK to add a list property to the script's output xml file.
What it basically does is add the real path (without the activatewindow and stuff) as a property to the menuitem.

This property value can be used to fill a dynamic list control.
Only "real" library paths are accepted including playlists.Also window properties and skin variables are accepted but they are only evaluated at script build not at runtime.

### DIFF
--- a/resources/lib/xmlfunctions.py
+++ b/resources/lib/xmlfunctions.py
@@ -597,10 +597,8 @@ class XMLFunctions():
                             newAction = newAction + "/" + actionPart
                     if len( actionParts ) == 2:
                         onclickelement.text = "ActivateWindow(" + actionParts[0] + "," + newAction + ")"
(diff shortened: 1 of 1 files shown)


ghost:
Dammit, I was just working on something like this which would produce the same result :) (Admittedly, I've not seen @HitcherUK's forum post, so my solution was adding a property to the Just Select Shortcut's method that returned without the extraneous stuff). Still, this seems to do the same (and works well from a very quick test) so I'll use it's implementation myself :). Thanks @marcelveldt.

BigNoid:
Looks good, thx!

marcelveldt:
good catch, thanks!

Issue #17, opened 2015-06-21
change shortcut deletes all properties


marcelveldt:
If a user wants to change the action for an existing shortcut all of the existing shortcut's properties like submenu, label, icon and stuff are lost.

I have received several complaints about this behaviour on my skin's forum threads so I thought looking into it.

@Ignoble61 Welcome back ;-) Hope you and your family are doing fine?
I was hoping if you could shine a light on this? Was this the expected behaviour?

If so, I'm thinking about adding an extra menu action to just change the action (and choose from the list) instead of wiping the full shortcut.

I can/will pick this one up myself (if nobody beats me to it), it's just a reminder that I need to do this and maybe someone else has an opinion about this.

ghost:
Hello @marcelveldt, nice to see you and, without going into too much detail, yes my family are doing fine and, indeed, I'm going to have three generations of my family living in one house for a couple of months whilst housing sales go through, which I'm both looking forward to an incredible amount, and dreading on every level ;) Hope you are keeping well, too, and congratulations on the birth of your offspring.

Wow, this is code I haven't looked at in a hell of a long time, so forgive me that everything I write is based on memory and not necessarily fact, but no - I don't think this is intended behaviour.

The problem is that when the action is changed, so is the labelID. The script keeps a list of labelID changes, and then - when the menu is saved - applies the previous additional properties from the previous labelID to the item with the new labelID. If my memory of all this is correct, then something's going wrong with either the saving of the previous labelID, or of the applying additional properties to the new item.

ghost:
Thinking off the top of my head (I'll take a proper look tomorrow, if no-one beats me too it) - since that code was written, the defaultID property was added - this may be a far better way of migrating properties from one item to another (an item with a new action is always considered a new item)...

ghost:
I'm unable to reproduce this, using Conq as a base test. When choosing a new shortcut via control 401, the action, label and thumbnail are all changed to that of the newly selected shortcut (expected behaviour).

The selected widget and background aren't changed, and it stays linked to the submenu.

marcelveldt:
Great to hear you're doing fine and glad to see you back here sharing your great ideas here with us.
Wow, that's a whole house full, can imagine that's fun in the beginning but very dreading in the longer term ;-)
Yes, I'm doing fine too. Enjoying every aspect of the newborn babygirl, it's a big change beiing a father but wouldn't give it up for all the money in the world :-)

[...]

PR #20, opened 2015-06-26
Isengard shortcuts

ghost:
Provides a version equivalency option to overrides.xml to check whether shortcuts/groupings with a <version/> tag apply to more than one version of Kodi. Makes it easier to manage shortcuts which change between Kodi versions. (Supersedes #15). Necessary because...

Switch available music library shortcuts to use Isengard Music Library Nodes. Also updates default shortcuts to use music library nodes on Isengard.

Two bugs that I came across whilst testing fixed also - invalid overrides.xml could lead to a build-loop, and defaultID set incorrectly on a new main menu when changing shortcut (may impact #17, but probably not).

### DIFF
--- a/resources/lib/datafunctions.py
+++ b/resources/lib/datafunctions.py
@@ -224,7 +224,7 @@ class DataFunctions():
             # Check that any version node matches current XBMC version
             version = node.find( "version" )
             if version is not None:
-                if __xbmcversion__ != version.text:
+                if __xbmcversion__ != version.text and self.checkVersionEquivalency( version.text, node.find( "action" ) ) == False:
                     tree.getroot().remove( node )
                     self._pop_labelID()
                     continue
@@ -397,7 +397,13 @@ class DataFunctions():
                 xbmcgui.Window( 10000 ).setProperty( "skinshortcuts-overrides-script-data", pickle.dumps( tree ) )
                 return tree
             except:
-                self._save_hash( overridepath, None )
+                if xbmcvfs.exists( overridepath ):
+                    # Unable to parse script overrides.xml
+                    log( "Unable to parse script overrides.xml. Invalid xml?" )
+                    self._save_hash( overridepath, xbmcvfs.File( overridePath ).read() )
+                else:
+                    # No script overrides.xml
+                    self._save_hash( overridepath, None )
                 xbmcgui.Window( 10000 ).setProperty( "skinshortcuts-overrides-script-data", "No overrides" )
                 return None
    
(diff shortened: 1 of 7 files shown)


ghost:
I'm not convinced the version equivalency is the best way to handle shortcuts changing between Kodi versions, but it's the best I can come up with. If anyone has any alternative ideas... :)

ghost:
I apologise for bumping a PR, but - as it appears no-one has any comments (and I presume, therefore, issues) regarding the version equivalency - is there anything I can do/change/whatever regarding this PR before it's pulled? I'd quite like to have the audio nodes in core for the code I'm working on currently :)

BigNoid:
I can give you commit rights to speed things up :)

ghost:
That's entirely up to you - though I'll be perfectly honest, it's been good to have my commits reviewed by a third party (Phil65 pointed out something that led to better code in #21, for example), rather than just making changes as I feel ;)

ghost:
@BigNoid I see you have given me commit rights - and many thanks. I'll do my best to ensure I don't abuse it (as I say, having code reviewed by a third party is a genuinely useful thing) - but I'll enjoy using it to merge this initially, and probably a few updates to templates (a feature I'm using extensively, but finding a little limited), which I was going to save until Kodi J\* - but I'll at least give people a change to notice my mistakes before I commit ;)

resources/lib/datafunctions.py
```diff
@@ -652,9 +671,59 @@ def checkVisibility ( self, action ):
             if path[ 1 ].endswith( ")" ):
                 path[ 1 ] = path[ 1 ][:-1]
             return NODE.get_visibility( path[ 1 ] )
-            
-        return ""
+
+        # Audio node visibility
+        elif action.startswith( "activatewindow(musiclibrary,musicdb://" ) or action.startswith( "activatewindow(10502,musicdb://" ) or action.startswith( "activatewindow(MusicLibrary,library://music/" ) or action.startswith( "activatewindow(10502,library://music/" ):
+            path = action.split( "," )
```
phil65: Is it safe to assume that every action contains a comma? If not then the next line would throw an exception.

ghost, line 677: [ignore me]

phil65:
sorry, missed this one. :)
Feel free to ping me if I dont answer within a few days.
Can only do some basic code checking though, not familiar with all that code. This one looks fine from what I can judge.

ghost:
The problem is I _am_ familiar with this code, and I really should have picked up on that. I'll ping you on the updated PR :)

ghost:
Looking at the code, it's absolutely safe to assume they include a comma - the comma is there in line 676 - `activatewindow(musiclibrary,musicdb://` (all of the possibilities that line is checking for include a comma...)

phil65:
yep, sorry, already noticed, comment was removed one minute later^^

ghost:
No worries, and nice to know I actually got it right :>

PR #21, opened 2015-06-28
Minor changes

ghost:
Some minor changes I've been meaning to make for a while...

Widget selection dialog will now wait for playlists to be loaded before displaying (previously playlists would not show if the user clicked Select Widget too quickly)

Add options to create shortcut for random and recursive slideshows from a pictures source

Handle errors saving the menu in the same way they are handled on a build menu failure - enable debug logging, recreate the error, disable debug logging, then either offer to upload a debug log or inform user of error. Previously, the window would simply refuse to close - rare scenario (and one I've never seen a debug log of to track down, except when I've broken things myself), and very unwanted behaviour.

(And two even more minor changes - changed string for log uploader to make it clearer it's an add-on that needs to be installed, and exactly what it's called, and fixed case where localised strings weren't being used in query to upload a log)

And that's it for changes I want to make to the script ATM, so PR's should quiet down for you guys ;)

### DIFF
--- a/resources/lib/gui.py
+++ b/resources/lib/gui.py
@@ -408,7 +408,72 @@ class GUI( xbmcgui.WindowXMLDialog ):
             # We set this to the default icon, so we need to check if /that/ icon is overriden
             self._get_icon_overrides( listitem, False )
         
-    def _save_shortcuts( self ):
+    def _save_shortcuts( self, weEnabledSystemDebug = False, weEnabledScriptDebug = False ):
+        # Entry point to save shortcuts - we will call the _save_shortcuts_function and, if it
+        # fails, enable debug options (if not enabled) + recreate the error, then offer to upload
+        # debug log (if relevant add-on is installed)
+
+        # Save the shortcuts
+        try:
+            self._save_shortcuts_function()
+            return
+        except:
+            print_exc()
+            log( "Failed to save shortcuts" )
+
+        # We failed to save the shortcuts
+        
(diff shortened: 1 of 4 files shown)


resources/lib/gui.py
```diff
@@ -1040,6 +1105,11 @@ def onClick(self, controlID):
                 
             # If playlists have been enabled for widgets, add them too
             if self.widgetPlaylists:
+                # Wait for playlists to be loaded
+                while LIBRARY.playlists() != True:
+                    xbmc.sleep( 100 )
```
phil65: are you sure that this will never hang here? (didnt look up that method)
Apart from that, all looks fine. :)

ghost:
As sure as I can be, and indeed doing a while != True is probably overkill.

LIBRARY.playlists() is the entry point for loading playlists. If they already loaded, it'll return True. If they're currently loading, it'll wait (without checking the code I think 2 seconds) then load them - then return True, or return True once they are loaded. If they're not currently loaded it'll load them and return True. The while != True is primarily just in case they won't load, which hopefully won't happen!

ghost:
Having reviewed, the != True is completely unnecessary - if it fails here, it'll fail in the main gui function - and it shouldn't in either (even if there's problems, the code should handle the error and return 0 playlists), so replaced with a simple call to the function, which will have the same effect with one line less of code ;)

phil65:
ok, great. Thx for your updates!

PR #25, opened 2015-07-07
Add ability for templates to get current text of property

ghost:
Changes to templates code.

Changes templates include to $SKINSHORTCUTS (from $skinshortcut - to better fit with $VAR, $INFO, etc).

Add ability for $SKINSHORTCUTS var to be set to the current value of a main menu item property.

@phil65 - I know you don't necessarily understand templates - and I know you don't use them - but would you mind casting a quick eye for any howlers before I pull this (if it helps, I'm relatively happy with this code ;)). Cheers.

### DIFF
--- a/resources/lib/template.py
+++ b/resources/lib/template.py
@@ -344,7 +344,10 @@ class Template():
                     continue
                     
                 # We've matched a property :)
-                properties[ name ] = property.text
+                if property.text:
+                    properties[ name ] = property.text
+                else:
+                    properties[ name ] = elem.text
                 break
         
         return properties
@@ -388,14 +391,14 @@ class Template():
                 tree.insert( index, newElement )
             
             # <tag>$skinshortcuts[var]</tag> -> <tag>[value]</tag>
-            if elem.text is not None and elem.text.startswith( "$skinshortcuts[" ) and elem.text[ 15:-1 ] in properties:
+            if elem.text is not None and elem.text.startswith( "$SKINSHORTCUTS[" ) and elem.text[ 15:-1 ] in properties:
                 # Replace the text with the property value
                 elem.text = properties[ elem.text[ 15:-1 ] ]
             
             # <tag attrib="$skinshortcuts[var]" /> -> <tag attrib="[value]" />
             for attrib in elem.attrib:
                 value = elem.attrib.get( attrib )
-                if value.startswith( "$skinshortcuts[" ) and value[ 15:-1 ] in properties:
+                if value.startswith( "$SKINSHORTCUTS[" ) and value[ 15:-1 ] in properties:
                     elem.set( attrib, properties[ value[ 15:-1 ] ] )
             
             # <skinshortcuts>visible</skinshortcuts> -> <visible>[condition]</visible>
--- a/resources/Templates.txt
+++ b/resources/Templates.txt
@@ -122,9 +122,9 @@ Property elements let you include a custom property in your template, dependant 
 [tag] - [Optional] The main menu must have an element with this tag
 [attributeName]|[attributeValue] - [Optional] The element must have an attribute with the name [attributeName] and the value [attributeValue]
 [textValue] - [Optional] The elment must have this value
-[propertyValue] - What you are setting the property to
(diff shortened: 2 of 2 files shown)


resources/lib/template.py
```diff
@@ -344,7 +344,10 @@ def getProperties( self, elem, items ):
                     continue
                     
                 # We've matched a property :)
-                properties[ name ] = property.text
+                if property.text is not None and property.text != "":
```
phil65: I prefer the more pythonic "if property.text:" in cases like that. probably matter of taste. :)

phil65:
looks fine. :)

resources/lib/template.py
ghost: I'd have to say it's definitely a matter of taste, especially in this case where changing it wouldn't actually simplify the code any (just re-arrange it so the 'false' cases come before the 'true' ones) - and would make it marginally harder to read for me! And when push comes to shove I'm not really a Python guy (indeed, after - according to the commits on this script - 2 years since I first worked with it, there's still things about Python I despise!)

If you feel strongly about it, though, do say and I'll do the re-arrangement :)

phil65: `if property.text:` definitely looks more readable to me than `if property.text is not None and property.text != "":` :)
Anyways, this is just cosmetics. Push the green button. ;)

ghost: Now, y'see, you put it like that and I understand ;) - I didn't get that I could them remove the `is not None` code - I really don't get Pythonic code :)

ghost:
@phil65 - hate to ask, but can you just double check I've got your Pythonic code correct? (And thanks for the heads up! Really must learn Pythonic coding at some point!)

resources/lib/template.py
phil65: you can do really crazy stuff in python. Not too long ago I found out that you can put a conditional into the class defintion:
`class ChildClass(ParentClass 1 if condition else ParentClass2)`
I am not sure yet if should find that elegant or ugly. :)

phil65:
All fine.
It´s still a matter of taste, the way I showed you is very implicit (it covers None, False, empty lists, empty strings etc), there are also people who dont like it that way. :)

ghost:
Thanks again. The problem is the only coding I ever actually learnt properly was Visual Basic 3. so it's no great surprise that a lot of my methodologies are ridiculously out of date! (Though in cases like this I think it's case of balancing between what I learned of making everything explicit against your 'pythonic' method where some things are implicit - though I instinctively know my methods are, by modern standards, very out of date!)

phil65:
yup, it probably depends on the coder´s background. I digged a lot into python during the last 6 months and I somehow feel that this language with all the duck-typing / unified protocols etc is very suited for implicit stuff. (last week I found some old code where I converted a string to a list of chars and processed that. Turned out that the exact same code also worked when I just kept it a string because iterating over it / indexing / merging is exactly the same for both. Same goes for lot of other types / classes)

resources/lib/template.py
ghost: I like explicit, so I'm going to file that as ugly for the moment ;)

ghost:
It probably doesn't help that I'm currently working on a Swift 1.x program at the moment, where explicit is not just preferred, it's absolutely essential. Perhaps it's not surprising that I've found Swift such an easy language to work with ;)

phil65:
But you probably have to admit that "if property.type:" looks way more sexy than the previous one. ;)

ghost:
Sexy? Sure! Readable? Depends what head I have on - if I have the one where I know that covers most eventualities then great (or, to put it another way, my 'Pythonic' head). If I have my more regular one where I know the two conditions I need to check for are None (or null, or whatever), or "", then not so much ;)

phil65:
True. But I can assure you based on empirics that "covers most cases" almost always translates to "covers all cases" :) And if not, then you will know it when writing that check.

ghost:
I get that, but my mind still struggles with the implicit coding. As I say, I know I need to learn more about it :) (It's times like this I realise just how old VB3 - and by extension I - am!)

PR #26, opened 2015-07-14
Rewrite of code to generate 'list' property for shortcuts

ghost:
Changes to code to generate 'list' property - supercedes #16, as discussed @14b215dd08
- Cleaner code, splitting on commas to grab appropriate part of the shortcut
- Only works for ActivateWindow shortcuts (other shortcuts will return as-is - let me know if there are any other shortcut types that should provide a specific 'list' property)
- Moved to datafunctions.py (because I intend to use this for updated widget selection code, and because xmlfunctions has become out of control and code needs moving out of there!)

@marcelveldt - this is your feature, so can you please review to ensure I haven't broken the use case you originally added this for.
@HitcherUK and @Nessus85100 - both of you reported issues with the original code, so can you please test to ensure the code fixes any issues / does not introduce any new issues / there's no other obvious use-cases I should add whilst I'm working on the code.

Cheers.

### DIFF
--- a/resources/lib/datafunctions.py
+++ b/resources/lib/datafunctions.py
@@ -939,6 +939,49 @@ class DataFunctions():
             text = text.replace('-', separator)
 
         return text
+
+    # ----------------------------------------------------------------
+    # --- Functions that should get their own module in the future ---
+    # --- (when xml building functions are revamped/simplified) ------
+    # ----------------------------------------------------------------
+
+    def getListProperty( self, onclick ):
+        # For ActivateWindow elements, extract the path property
+        if onclick.startswith( "ActivateWindow" ):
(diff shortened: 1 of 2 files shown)


marcelveldt:
@Ignoble61, It looks fine to me. Great thing too about changing the widget selection, that is something that is really wanted/needed.
atm I used a very complicated workaround for that in my skin, would be good if that could be replaced by this.

This request btw came from @HitcherUK Maybe he can tell us a little more about the use case he wants to use this list property for. I suspect it has something to do with choosing widget content. In that case the new widget selector would be even better.

marcelveldt:
code working fine on my testsetup

ghost:
Thanks for testing + merging :+1:

Nessus85100:
Also everything working fine for me too. The missing bracket it's fixed.

PR #29, opened 2015-07-18
Speed up menu build

ghost:
Changes to menu build so that each menu item is only built once, then has visibility conditions applied to a copy (rather than building menu items multiple times with varying visibility conditions.) In my own tests I've seen speeds increase by a little over half.

@marcelveldt As you've highlighted build speed issues (and because, fwiu, Titan uses most every build option there is, including building all menus in a single list) can you test for speed improvements/lack of breakages please. Cheers.

### DIFF
--- a/resources/lib/xmlfunctions.py
+++ b/resources/lib/xmlfunctions.py
@@ -3,6 +3,7 @@ import os, sys, datetime, unicodedata, re
 import xbmc, xbmcgui, xbmcvfs, xbmcaddon
 import xml.etree.ElementTree as xmltree
 from xml.sax.saxutils import escape as escapeXML
+import copy
 from traceback import print_exc
 
 if sys.version_info < (2, 7):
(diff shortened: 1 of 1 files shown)


marcelveldt:
I will do a complete test within a few days. Great stuff btw, maybe with this speed fixes I can go back to using overrides for skinshortcuts again instead of setting skin strings for the backgrounds.

ghost:
If you find it's still unreasonably slow, get me a couple of profiling logs (uncomment the profiling code at the bottom of default.py - remembering to comment out the standard call!) and let me know if there are any particular overrides that bog it down, or if its cumulative. Not saying there's much more I can do, but I'll be happy to see if there are other speed improvements that can be got.

marcelveldt:
OK, I tested this change on my testsetup. At first, the menu build seem to be a bit faster. Just a few seconds but it's noticeable. Also I compared the script's output and there is nothing missing. So far so good to continue with this merge.

But... As a real test I decided to put back my old overrides.xml which contained lot's of defaultvalues, like widgetdefault and backgrounddefault. Back then that was the whole reason the script was to slow and I decided to store the backgrounds and widgets in a skin string instead of using skinshortcuts for it.

So, I reverted to my overrides.xml which still contained all the widget and background stuff. Result: Really slow building of the menus. On my Core I5 testmachine it takes almost a minute to build the menu. On a slow device like a pi it would take 20 minutes or so (that's how it was back then).

For me it's no big deal because my workaround with storing the default background and widget in a skin string works okay for now but maybe this is a problem for others who also have defaults set in their overrides.xml. I can imagine btw that providing the defaults would be slow because all items have to compared every single time so leave it if it can't be fixed.

The log  --> http://pastebin.com/cRpyMAew

ghost:
Thanks for the profiling log - that showed one huge time waster (as it were) which I've now fixed on the build-rewrite branch (https://github.com/Ignoble61/script.skinshortcuts/tree/build-rewrite) - from my tests its showing a huge improvement by changing how the overrides.xml are loaded (changing from trying to keep from reloading between runs with a lot of time being wasted in un-pickling them, to just keeping them in memory for a single run). With Titan Beta my build times are going from approx 16 seconds to approx 3! Do you mind testing and seeing if it speeds things up for you? Cheers.

(It shows a couple of other areas where improvements may be got, so I'll take a look at those areas another day before doing another PR, but no improvements should be as big as the overrides.xml loading one)

marcelveldt:
Wow! I just tested with your cleanup branch... Well, that's impressing :-)
The build is now less then a second on my machine, even with the whole bunch of overrides added.
Also, I can confirm (so far) that nothing broke. It all seems to be working very well.

If you want I can publish a beta version of the script on the emby beta repo I use for my skin, that will allow us for a handful of testers to check if there are any issues before all changes go to the Kodi repo.

marcelveldt:
btw: feel free to merge this changes in the main branch. As far as I can see all is well.

ghost:
The changes should be fairly safe, so I'll tidy up & merge when I get time :)

PR #31, opened 2015-07-21
Fixed UnicodeWarning in comparison of translated <None> label

roee88:
A fix for the bug described in http://forum.kodi.tv/showthread.php?tid=178294&pid=2058442#pid2058442

Empty sub-menus should normally be ignored. 
When using some translations the comparison fails and so there is no way to remove the empty sub-menu.

Reason: the translated <None> text is using unicode while the label is a utf-8 encoded string.


marcelveldt:
Thanks for the fix. Looking fine.

PR #30, opened 2015-07-18
Remove old functions

ghost:
Removes all .shortcut file compatibility (should only affect Nox5 + variants)
Removes ability for list to be filled with controls to launch management dialog (I don't believe these methods are in use by any skin)

@BigNoid - as these changes should only affect your skin + variants, do you want the honour of merging when you feel comfortable (idk if you prefer to wait for your change to .DATA.xml files - #22 - to hit the repo before this gets merged)?

PR #33, opened 2015-07-28
Additional unicode fixes

roee88:
Fixes for custom thumbnail,icon,background paths with unicode characters.

### DIFF
--- a/resources/lib/datafunctions.py
+++ b/resources/lib/datafunctions.py
@@ -8,6 +8,7 @@ from xml.dom.minidom import parse
 from traceback import print_exc
 from htmlentitydefs import name2codepoint
 from unidecode import unidecode
+from unicodeutils import try_decode
 
 import nodefunctions
 NODE = nodefunctions.NodeFunctions()
@@ -137,10 +138,7 @@ class DataFunctions():
             paths = [userShortcuts, skinShortcuts, defaultShortcuts ]
         
         for path in paths:
-            try:
-                path = path.decode( "utf-8" )
-            except:
-                pass
+            path = try_decode( path )
                 
             tree = None
             altPath = path.replace( ".DATA.xml", ".shortcuts" )
@@ -809,10 +807,7 @@ class DataFunctions():
         if data is None:
             return ["","","",""]
         
-        try:
-            data = data.decode( "utf-8" )
-        except:
-            pass
+        data = try_decode( data )
 
         skinid = None
         lasttranslation = None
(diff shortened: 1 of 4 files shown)


resources/lib/gui.py
```diff
@@ -473,6 +473,12 @@ def _save_shortcuts( self, weEnabledSystemDebug = False, weEnabledScriptDebug =
                 # Inform user menu couldn't be saved
                 xbmcgui.Dialog().ok( __addon__.getAddonInfo( "name" ), __language__( 32097 ), __language__( 32094 ) )                 
 
+    def _try_decode(self, text, encoding="utf-8"):
+        try:
+            return text.decode(encoding)
+        except:
+            return text
```
phil65: it would probably make more sense to put this into a separate module instead of adding it to gui class?

ghost:
Not sure I agree with adding a module for a single function (but it's not rare for phil65 and I to disagree on coding methodologies ;)), otherwise the changes look good. @phil65, if you're happy with the new module then I think this can be merged.

And thanks for the fixes.

resources/lib/datafunctions.py
```diff
@@ -1107,10 +1102,8 @@ def upgrade_xmlfile( self, path, mixedVersion = False, saveFile = True ):
             actionTree = xmltree.SubElement( root, "shortcut" )
             
             # Action
-            try:
-                action = urllib.unquote( shortcut[4] ).decode( "utf-8" )
-            except:
-                action = urllib.unquote( shortcut[4] )
+            action = try_decode( urllib.unquote( shortcut[4] ) )
```
ghost, line 1105: Just a note, the upgrade_\* functions will be dropped when #30 is merged :)

phil65:
I doesnt have to be a new module, but since that method does not really do anything concerning the state of the class (=it doesnt need self) it should either be static or get moved out to some module with helper functions.

ghost:
Now there we agree ;) My instinct was it would fit nicely in datafunctions.py (which actually isn't static/a static class/a module because this script is really badly coded on so many levels, but it at least does have a lot of helper functions), which every other class/module/whatever has an instance of, and is used for bringing most data into the script and checking it - possibly somewhere around the local() code, which is very broadly similar.

It's also good to see how moving it out of gui.py meant it could be used in so many more situations :)

Edit: Actually this might be a good opportunity to lay the groundwork as it were for moving functions out of classes and into modules, if instead this function was in a more generically named module. There's a lot of code, particularly in datafunctions.py, which could very easily be - and in the long-run SHOULD be - a function in a module, not a class.

marcelveldt:
@Ignoble61 will you do the merge if everything is fine here ?

ghost:
Fundamentals here are great, and the try_decode() can be moved somewhere more suitable (or unicodeutils can be be made more useful) in the future, so let's get this merged. Thanks again for the fixes, @roee88

PR #37, opened 2015-08-08
Add button 312 for new selection of widgets

ghost:
Incomplete list of changes included:-
- GUI button 312 added to select widgets including custom widgets, library nodes, sources, add-ons, playlists
- Deprecate GUI button 309 (old method to select widget)
- All the code needed for the new widgets to return 5 properties - widget, widgetName, widgetType, widgetTarget, widgetPath
- New widgetdefaultnode override
- Support for additional custom groupings of available shortcuts for Just Select method + GUI buttons 401 & 307
- Remove choice of action in select dialog when using widget groupings (will always return ActivateWindow path rather than giving choice of, for example, starting a slideshow)
- Fixes for browsing/selecting audio nodes (notably, correct window returned, fix potential crash, fix log errors)
- Move code to load skin-defined widgets to library.py
- Allow exploring of plugin elements provided by program add-ons
- Add ability to go 'back' when selecting a shortcut/widget
- Template improvements - ability to match any condition rather than all, more fine control of getting/setting properties, ability to use an include in a property, ability to build multiple Other templates per main menu item, minor fixes
- Fix security vulnerability in .hash files (arbitrary python code could be run)
- Minor fixes

Thanks to @marcelveldt and @Nessus85100 for testing, and marcelveldt for the code letting add-ons return the new widgetType property.

Marcelveldt, I know it's a lot of code, but do you fancy a code review :grinning:

### DIFF
--- a/resources/lib/datafunctions.py
+++ b/resources/lib/datafunctions.py
@@ -474,48 +474,84 @@ class DataFunctions():
         # Load skin defaults (in case we need them...)
         tree = self._get_overrides_skin()
         if tree is not None:
-            for elemSearch in [["widget", tree.findall( "widgetdefault" )], ["background", tree.findall( "backgrounddefault" )], ["custom", tree.findall( "propertydefault" )] ]:
+            for elemSearch in [["widget", tree.findall( "widgetdefault" )], ["widget:node", tree.findall( "widgetdefaultnode" )], ["background", tree.findall( "backgrounddefault" )], ["custom", tree.findall( "propertydefault" )] ]:
                 for elem in elemSearch[1]:
+                    # Get labelID and defaultID
+                    labelID = elem.attrib.get( "labelID" )
+                    defaultID = labelID
+                    if "defaultID" in elem.attrib:
+                        defaultID = elem.attrib.get( "defaultID" )
+
                     if elemSearch[0] == "custom":
                         # Custom property
                         if "group" not in elem.attrib:
-                            defaultProperties.append( ["mainmenu", elem.attrib.get( 'labelID' ), elem.attrib.get( 'property' ), elem.text, elem.attrib.get( 'defaultID' ) ] )
+                            defaultProperties.append( ["mainmenu", labelID, elem.attrib.get( 'property' ), elem.text, defaultID ] )
                         else:
-                            defaultProperties.append( [elem.attrib.get( "group" ), elem.attrib.get( 'labelID' ), elem.attrib.get( 'property' ), elem.text, elem.attrib.get( 'defaultID' ) ] )
+                            defaultProperties.append( [elem.attrib.get( "group" ), labelID, elem.attrib.get( 'property' ), elem.text, defaultID ] )
                     else:
                         # Widget or background
                         if "group" not in elem.attrib:
-                            defaultProperties.append( [ "mainmenu", elem.attrib.get( 'labelID' ), elemSearch[0], elem.text, elem.attrib.get( 'defaultID' ) ] )
+                            defaultProperties.append( [ "mainmenu", labelID, elemSearch[ 0 ].split( ":" )[ 0 ], elem.text, defaultID ] )
                             
                             if elemSearch[ 0 ] == "background":
                                 # Get and set the background name
                                 backgroundName = self._getBackgroundName( elem.text )
                                 if backgroundName is not None:
-                                    defaultProperties.append( [ "mainmenu", elem.attrib.get( "labelID" ), "backgroundName", backgroundName, elem.attrib.get( 'defaultID' ) ] )
+                                    defaultProperties.append( [ "mainmenu", labelID, "backgroundName", backgroundName, defaultID ] )
                                 
                             if elemSearch[0] == "widget":
                                 # Get and set widget type and name
                                 widgetDetails = self._getWidgetNameAndType( elem.text )
                                 if widgetDetails is not None:
-                                    defaultProperties.append( [ "mainmenu", elem.attrib.get( "labelID" ), "widgetName", widgetDetails[0], elem.attrib.get( 'defaultID' ) ] )
-                                    if widgetDetails[1] is not None:
-                                        defaultProperties.append( [ "mainmenu", elem.attrib.get( "labelID" ), "widgetType", widgetDetails[1], elem.attrib.get( 'defaultID' ) ] )
+                                    defaultProperties.append( [ "mainmenu", labelID, "widgetName", widgetDetails[ "name" ], defaultID ] )
+                                    if "type" in widgetDetails:
+                                        defaultProperties.append( [ "mainmenu", labelID, "widgetType", widgetDetails[ "type" ], defaultID ] )
+                                    if "path" in widgetDetails:
+                                        defaultProperties.append( [ "mainmenu", labelID, "widgetPath", widgetDetails[ "path" ], defaultID ] )
+                                    if "target" in widgetDetails:
+                                        defaultProperties.append( [ "mainmenu", labelID, "widgetTarget", widgetDetails[ "target" ], defaultID ] )
+
+                            if elemSearch[0] == "widget:node":
+                                # Set all widget properties from the default
+                                if "label" in elem.attrib:
+                                    defaultProperties.append( [ "mainmenu", labelID, "widgetName", elem.attrib.get( "label" ), defaultID ] )
+                                if "type" in elem.attrib:
+                                    defaultProperties.append( [ "mainmenu", labelID, "widgetType", elem.attrib.get( "type" ), defaultID ] )
+                                if "path" in elem.attrib:
+                                    defaultProperties.append( [ "mainmenu", labelID, "widgetPath", elem.attrib.get( "path" ), defaultID ] )
+                                if "target" in elem.attrib:
+                                    defaultProperties.append( [ "mainmenu", labelID, "widgetTarget", elem.attrib.get( "target" ), defaultID ] )
                         else:
-                            defaultProperties.append( [ elem.attrib.get( "group" ), elem.attrib.get( 'labelID' ), elemSearch[0], elem.text, elem.attrib.get( 'defaultID' ) ] )
+                            defaultProperties.append( [ elem.attrib.get( "group" ), labelID, elemSearch[ 0 ].split( ":" )[ 0 ], elem.text, defaultID ] )
                             
                             if elemSearch[ 0 ] == "background":
                                 # Get and set the background name
                                 backgroundName = self._getBackgroundName( elem.text )
                                 if backgroundName is not None:
-                                    defaultProperties.append( [ "mainmenu", elem.attrib.get( "labelID" ), "backgroundName", backgroundName, elem.attrib.get( 'defaultID' ) ] )
+                                    defaultProperties.append( [ "mainmenu", labelID, "backgroundName", backgroundName, defaultID ] )
                             
                             if elemSearch[0] == "widget":
                                 # Get and set widget type and name
                                 widgetDetails = self._getWidgetNameAndType( elem.text )
                                 if widgetDetails is not None:
-                                    defaultProperties.append( [ elem.attrib.get( "group" ), elem.attrib.get( "labelID" ), "widgetName", widgetDetails[0], elem.attrib.get( 'defaultID' ) ] )
-                                    if widgetDetails[1] is not None:
-                                        defaultProperties.append( [ elem.attrib.get( "group" ), elem.attrib.get( "labelID" ), "widgetType", widgetDetails[1], elem.attrib.get( 'defaultID' ) ] )                
+                                    defaultProperties.append( [ elem.attrib.get( "group" ), labelID, "widgetName", widgetDetails[ "name" ], defaultID ] )
+                                    if "type" in widgetDetails:
+                                        defaultProperties.append( [ elem.attrib.get( "group" ), labelID, "widgetType", widgetDetails[ "type" ], defaultID ] )
+                                    if "path" in widgetDetails:
+                                        defaultProperties.append( [ elem.attrib.get( "group" ), labelID, "widgetPath", widgetDetails[ "path" ], defaultID ] )
+                                    if "target" in widgetDetails:
+                                        defaultProperties.append( [ elem.attrib.get( "group" ), labelID, "widgetTarget", widgetDetails[ "target" ], defaultID ] )
+
+                            if elemSearch[ 0 ] == "widget:node":
+                                # Set all widget properties from the default
+                                if "label" in elem.attrib:
+                                    defaultProperties.append( [ elem.attrib.get( "group" ), labelID, "widgetName", elem.attrib.get( "label" ), defaultID ] )
+                                if "type" in elem.attrib:
+                                    defaultProperties.append( [ elem.attrib.get( "group" ), labelID, "widgetType", elem.attrib.get( "type" ), defaultID ] )
+                                if "path" in elem.attrib:
+                                    defaultProperties.append( [ elem.attrib.get( "group" ), labelID, "widgetPath", elem.attrib.get( "path" ), defaultID ] )
+                                if "target" in elem.attrib:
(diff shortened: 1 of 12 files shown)


ghost:
I feel cheeky asking this, but @phil65 - again, I know it's an awful lot of code, but if you have the time to give it a quick look over for any obvious mistakes, that would be hugely appreciated (and hugely understood if you don't have time :wink:)

marcelveldt:
I have reviewed the code, it's indeed a lot of code and I couldn't find any "easy to spot" errors. Also I have tested with your branch and the results are fantastic but I've already let you know.
I think merging the code is safe to do but I suggest to hold off with a update to official Kodi repo. My suggestion is to first build a beta version to let some users test it. If all is fine publish the update to official Kodi repo.

The beta version can be included on the beta repo of my skin. Not sure if any other skinners also have a beta repo or are willing to provide a beta version of the script with their (beta) skins...

What do you think ?

ghost:
Thanks for taking the time to look through it all :) Additional testing sounds eminently sensible if you (or others) want to include it on a beta repo.

Nessus85100:
I could also put this on beta testing with Bello but my Isengard version is not ready yet and i prefer not do it in Helix. If this is still on beta stage when Bello Isengard is ready, i will include it for testing.

Cheers
Nessus

resources/lib/datafunctions.py
```diff
@@ -474,48 +474,84 @@ def _get_additionalproperties( self, profileDir ):
...
+                            if elemSearch[ 0 ] == "widget:node":
+                                # Set all widget properties from the default
+                                if "label" in elem.attrib:
+                                    defaultProperties.append( [ elem.attrib.get( "group" ), labelID, "widgetName", elem.attrib.get( "label" ), defaultID ] )
+                                if "type" in elem.attrib:
+                                    defaultProperties.append( [ elem.attrib.get( "group" ), labelID, "widgetType", elem.attrib.get( "type" ), defaultID ] )
+                                if "path" in elem.attrib:
+                                    defaultProperties.append( [ elem.attrib.get( "group" ), labelID, "widgetPath", elem.attrib.get( "path" ), defaultID ] )
+                                if "target" in elem.attrib:
+                                    defaultProperties.append( [ elem.attrib.get( "group" ), labelID, "widgetTarget", elem.attrib.get( "target" ), defaultID ] )
```
phil65, line 555: I always prefer something like this for all kind of repititions:

``` python
dct = {"label": "widgetName",
       ....}
for key, value in dct.items():
    if key in elem.attrib:
        defaultProperties.append(....)
```

(either using dict or using a list of tuples)

resources/lib/datafunctions.py
```diff
@@ -474,48 +474,84 @@ def _get_additionalproperties( self, profileDir ):
...
+                                    defaultProperties.append( [ "mainmenu", labelID, "widgetName", elem.attrib.get( "label" ), defaultID ] )
+                                if "type" in elem.attrib:
+                                    defaultProperties.append( [ "mainmenu", labelID, "widgetType", elem.attrib.get( "type" ), defaultID ] )
+                                if "path" in elem.attrib:
+                                    defaultProperties.append( [ "mainmenu", labelID, "widgetPath", elem.attrib.get( "path" ), defaultID ] )
+                                if "target" in elem.attrib:
+                                    defaultProperties.append( [ "mainmenu", labelID, "widgetTarget", elem.attrib.get( "target" ), defaultID ] )
                         else:
-                            defaultProperties.append( [ elem.attrib.get( "group" ), elem.attrib.get( 'labelID' ), elemSearch[0], elem.text, elem.attrib.get( 'defaultID' ) ] )
+                            defaultProperties.append( [ elem.attrib.get( "group" ), labelID, elemSearch[ 0 ].split( ":" )[ 0 ], elem.text, defaultID ] )
```
phil65, line 526: is it safe to say that accessing those list elements does not throw an exception? ( elemSearch[ 0 ].split( ":" )[ 0 ] )

resources/lib/gui.py
```diff
@@ -992,6 +983,7 @@ def onClick(self, controlID):
         if controlID == 307:
             # Change Action
             log( "Change action (307)" )
+            currentWindow = xbmcgui.Window( xbmcgui.getCurrentWindowDialogId() )
```
phil65: I always assign self.window_id in onInit() already since for most WindowXML instances it is used pretty often in all kind of methods.

resources/lib/library.py
```diff
@@ -88,11 +88,12 @@ def __init__( self, *args, **kwargs ):
...
         # Empty dictionary for different shortcut types
-        self.dictionaryGroupings = {"common":None, "commands":None, "video":None, "movie":None, "movie-flat":None, "tvshow":None, "tvshow-flat":None, "musicvideo":None, "musicvideo-flat":None, "customvideonode":None, "customvideonode-flat":None, "videosources":None, "pvr":None, "radio":None, "pvr-tv":None, "pvr-radio":None, "music":None, "musicsources":None, "picturesources":None, "playlist-video":None, "playlist-audio":None, "addon-program":None, "addon-video":None, "addon-audio":None, "addon-image":None, "favourite":None, "settings":None }
+        self.dictionaryGroupings = {"common":None, "commands":None, "video":None, "movie":None, "movie-flat":None, "tvshow":None, "tvshow-flat":None, "musicvideo":None, "musicvideo-flat":None, "customvideonode":None, "customvideonode-flat":None, "videosources":None, "pvr":None, "radio":None, "pvr-tv":None, "pvr-radio":None, "music":None, "musicsources":None, "picturesources":None, "playlist-video":None, "playlist-audio":None, "addon-program":None, "addon-video":None, "addon-audio":None, "addon-image":None, "favourite":None, "settings":None, "widgets":None, "widgets-classic":[] }
```
phil65: using multiple lines would probably make sense here :)

resources/lib/datafunctions.py
ghost, line 555: Yes, that's a better method, but not sure it's useful here - for the widget:node's, the values are being accessed directly from an elementtree object. To do as you suggest, I'd have to do what's being done here (checking whether the element has the attribute) and assign it to a dict, then assign from that dict to the defaultProperties.

resources/lib/library.py
```diff
@@ -238,6 +263,27 @@ def retrieveContent( self, content ):
             items = self.checkForFolder( items )
         else:
             items = []
+
+        # Add widget information for video/audio nodes
+        if content == "video" or content == "music":
```
phil65: I always prefer "if content in ["video", "music"]" for stuff like this.

resources/lib/datafunctions.py
ghost, line 526: Yes, it is ;) We know that elemSearch[ 0 ] will never be empty (its various values are set slightly higher up). If it doesn't have a ':' in, then it gets 'split' into a list with just a single entry [ 0 ]

resources/lib/gui.py
ghost: Very true. Will update, cheers.

resources/lib/library.py
```diff
@@ -238,6 +263,27 @@ def retrieveContent( self, content ):
...
+            # Video nodes - add widget information
+            for listitem in items:
+                path = listitem.getProperty( "path" )
+                if path.lower().startswith( "activatewindow" ):
```
phil65, line 299: ...and
`if not path.lower().startswith( "activatewindow" ): continue`for these cases to save some levels of indentation. Probably matter of taste though.

resources/lib/library.py
```diff
@@ -1279,12 +1345,46 @@ def addons( self ):
...
+                                        otherItem = self._create([path, item['name'] + "  >", "::SCRIPT::32012", {"icon": "DefaultAddon.png", "thumb": thumb} ])
+                                        otherItem.setProperty( "path", "||BROWSE||" + item['addonid'].encode('utf-8') )
+                                        otherItem.setProperty( "action", "RunAddOn(" + item['addonid'].encode('utf-8') + ")" )
+                                        imageItems[ item[ "name" ] ] = otherItem
+                                    elif content == "executable":
+                                        otherItem = self._create([path, item['name'] + "  >", shortcutType, {"icon": "DefaultAddon.png", "thumb": thumb} ])
+                                        otherItem.setProperty( "path", "||BROWSE||" + item['addonid'].encode('utf-8') )
+                                        otherItem.setProperty( "action", "RunAddOn(" + item['addonid'].encode('utf-8') + ")" )
+                                        listitems[ item[ "name" ] + ".1" ] = otherItem
+
```
phil65, line 1354: this block probably could also get done with less and non-repeating code by using a list of tuples / dict?

resources/lib/library.py
```diff
@@ -1305,6 +1405,122 @@ def addons( self ):
...
+            ##### VIDEO ITEMS ####
+            if (item["showtitle"] and not item["artist"]):
+                #this is a tvshow, episode or season...
+                if (item["season"] > -1 and item["episode"] == -1):
+                    return "seasons"
+                elif item["season"] > -1 and item["episode"] > -1:
+                    return "episodes"
+                else:
+                    return "tvshows"
+            elif (item["artist"]):
```
phil65, line 1425: several unneeded brackets in this block. (now I am gettin nitpicky, hehe :) )

ghost, line 1425: Get as nit-picky as you like with that function - I didn't write that one ;)

resources/lib/library.py
```diff
@@ -1305,6 +1405,122 @@ def addons( self ):
...
+                #this is a musicvideo!
+                return "musicvideos"
+            elif (item["imdbnumber"] or item["mpaa"] or item["trailer"] or item["studio"]):
+                return "movies"
+
+        return None
+
+    def widgets( self ):
+        if self.loadedWidgets == True:
```
phil65: since "True" and "False" are singletons it is recommended to use "is True" or just "if self.loadedWidgets:"
see here for example for bit more details: https://stackoverflow.com/questions/27276610/boolean-identity-true-vs-is-true

resources/lib/library.py
```diff
@@ -1305,6 +1405,122 @@ def addons( self ):
...
+            # The list has already been populated
+            return self.loadedWidgets
+        elif self.loadedWidgets == "Loading":
+            # The list is currently being populated, wait and then return it
+            count = 0
+            while count < 20:
```
phil65: for i in range(0,20) is probably more pythonic :)

resources/lib/library.py
```diff
@@ -1305,6 +1405,122 @@ def addons( self ):
...
+            # We're going to populat the list
+            self.loadedWidgets = "Loading"
+
+        # Load skin overrides
+        tree = DATA._get_overrides_skin()
+
+        # Get widgets
+        listitems = []
+        log( "Loading skin widgets" )
+        try:
```
phil65: perhaps nicer to only put the stuff into try-except which really can fail?

ghost: He he, yup I'll look again at the widget() function (and pretty much all the library loading functions, which I copied/pasted and stuck the widget code into!)

resources/lib/library.py
```diff
@@ -1316,52 +1532,91 @@ def sortDictionary( self, dictionary ):
...
+        if len( label ) is not 1:
+            dialogLabel = label[0].replace( "  >", "" ) + " - " + label[ len( label ) - 1 ].replace( "  >", "" )
 
-        # Default action - create shortcut
         listings = []
         
         tree = DATA._get_overrides_skin()
+
+        # Shortcut to go 'up'
+        if len( label ) is 1:
```
phil65: for comparing with numbers i would use == and != (even though -5 to 255 are singletons, but changing the number to something outside that range would break this function when using "is")

resources/lib/library.py
```diff
@@ -1683,51 +2008,75 @@ def selectShortcut( self, group = "", custom = False, availableShortcuts = None,
...
+                    # Return actionShow as chosenPath
                     selectedShortcut.setProperty( "chosenPath", selectedShortcut.getProperty( "action-show" ) )
                 else:
-                    selectedShortcut.setProperty( "chosenPath", selectedShortcut.getProperty( "action-play" ) )
+                    # Give the user the choice of playing or displaying the playlist
+                    dialog = xbmcgui.Dialog()
+                    userchoice = dialog.yesno( __language__( 32040 ), __language__( 32060 ), "", "", __language__( 32061 ), __language__( 32062 ) )
+                    # False: Display
+                    # True: Play
+                    if userchoice == False:
```
phil65: "False" is singleton so "is" or just "if userchoice:" is preferred.

resources/lib/nodefunctions.py
```diff
@@ -202,6 +225,61 @@ def get_visibility( self, path ):
...
+            defaultPath = path.replace( "library://video", os.path.join( xbmc.translatePath( "special://xbmc".decode('utf-8') ), "system", "library", "video" ) ) + "index.xml"
+            defaultFile = path.replace( "library://video", os.path.join( xbmc.translatePath( "special://xbmc".decode('utf-8') ), "system", "library", "video" ) )[:-1] + ".xml"
+        elif "library://music" in path:
+            customPath = path.replace( "library://music", os.path.join( xbmc.translatePath( "special://profile".decode('utf-8') ), "library", "music" ) ) + "index.xml"
+            customFile = path.replace( "library://music", os.path.join( xbmc.translatePath( "special://profile".decode('utf-8') ), "library", "music" ) )[:-1] + ".xml"
+            defaultPath = path.replace( "library://music", os.path.join( xbmc.translatePath( "special://xbmc".decode('utf-8') ), "system", "library", "music" ) ) + "index.xml"
+            defaultFile = path.replace( "library://music", os.path.join( xbmc.translatePath( "special://xbmc".decode('utf-8') ), "system", "library", "music" ) )[:-1] + ".xml"
+        else:
```
phil65, line 244: also lot of repeating code here I think? Perhaps worth refactoring it a bit.

resources/lib/template.py
```diff
@@ -203,20 +206,47 @@ def findSubmenu( self, name, level ):
...
             template = copy.deepcopy( elem )
-            match = True
+            matched = True
+
+            # Check whether the skinner has set the match type (whether all conditions need to match, or any)
+            matchType = "all"
+            matchElem = template.find( "match" )
+            if matchElem is not None:
+                matchType = matchElem.text.lower()
+                if matchType != "any" and matchType != "all":
```
phil65: if matchType not in ["any", "all"]:

resources/lib/template.py
```diff
@@ -203,20 +206,47 @@ def findSubmenu( self, name, level ):
...
+                    log( "Invalid <match /> element in template" )
+                    matchType = "all"
+                if matchType == "any":
```
phil65: elif would make very bit more sense :)

phil65:
some remarks, nothing serious.
Only a superficial review though, I dont know the code well enough for a real proper one.

ghost:
That'll teach me to ask you to look at my code :stuck_out_tongue: Seriously, though, thanks so much for taking the time to read through and comment. Hugely appreciated. There are a small number there I disagree with (primarily due just to my coding style I think), but the vast majority are damn good insights, and will keep me busy for a couple of hours at least. :)

resources/lib/library.py
marcelveldt, line 1425: Magic code :-)

ghost:
That should be most of them sorted. Cheers again, @phil65

phil65:
no problem. As said, all are basically just "recommendations", so if you disagree with some then it´s fine of course to just ignore them.

ghost:
So, with those fixes in, everyone happy for this to be merged?

marcelveldt:
Yep, happy with the merge. Thanks for your incredible hard work on this once again. As soon as the merge is done I'll build a beta package. Thanks @phil65 for reviewing!

marcelveldt:
@Nessus85100 if you want to include the beta on your beta repo, here's the zipfile:

https://github.com/BigNoid/script.skinshortcuts/archive/0.5.1.zip

Nessus85100:
Thanks. Is this any different from the .zip download in first page ?

marcelveldt:
Yes, it's a specific version. That way you can track errors that are in a specific release.

PR #41, opened 2015-08-11
Minor fixes to widgets and others

### LINKED ISSUE #40: Script's main window multiple instances
Accidentally the other day i push the button to open script's main window twice and i notice that the window was open twice. Tested again and it turns out that the window is opening so many times how many times the window opens so many times as the times you manage to click before the first instance is open. I manage to do that for 6 times in my testing.

Is this a bug or the multiple instances have some purpose ?

Cheers
Nessus

### LINKED ISSUE #42: Issue with widget-groupings
I am pretty sure that i am doing something wrong here but i have a problem with widget-groupings and the new way of assigning widgets. I am trying to use the `<widget-groupings>` so to group all of my skin's widgets to custom nodes. The problem is that the `widgetName` property is always empty for my skin's widgets. When i assign a widget from the default libraries nodes they are showing fine. Also for some weird reason the Smartish script widgets are not showing in the select dialog (the script is installed fine of course).

This is my `<widget-groupings>` code... http://pastebin.com/g4mPPeQs

What am i doing wrong here ?

ghost:
Widget fixes:-
- Fix widgetName for `<shortcut />` elements (closes #42)
- Fix potential script error when selecting GUI 309 or 312
- Fix indentical templates in different includes were being merged

Other fixes
- Remove visibility condition from link to music addons (as reported on the forum)
- Don't allow multiple management dialogs to be loading at the same time (closes #40)

### DIFF
--- a/default.py
+++ b/default.py
@@ -237,20 +237,25 @@ class Main:
                     xbmc.executebuiltin( singleAction )
         
     
-    def _manage_shortcuts( self, group, defaultGroup, nolabels, groupname ):            
+    def _manage_shortcuts( self, group, defaultGroup, nolabels, groupname ):
+        homeWindow = xbmcgui.Window( 10000 )
+        if homeWindow.getProperty( "skinshortcuts-loading" ):
+            return
+
+        homeWindow.setProperty( "skinshortcuts-loading", "True" )
         import gui
         ui= gui.GUI( "script-skinshortcuts.xml", __cwd__, "default", group=group, defaultGroup=defaultGroup, nolabels=nolabels, groupname=groupname )
         ui.doModal()
         del ui
         
         # Update home window property (used to automatically refresh type=settings)
-        xbmcgui.Window( 10000 ).setProperty( "skinshortcuts",strftime( "%Y%m%d%H%M%S",gmtime() ) )
+        homeWindow.setProperty( "skinshortcuts",strftime( "%Y%m%d%H%M%S",gmtime() ) )
         
         # Clear window properties for this group, and for backgrounds, widgets, properties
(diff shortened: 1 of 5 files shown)


ghost:
@marcelveldt thanks for reporting the fixes on the forum. It's much appreciated :)

If it helps, I'll put money on the build error also reported on the forums on the user in question using some exotic path substitution like substituting their userdata folder (technically supported by Kodi, but basically unsupported by Kodi Addons) - the error looks like their profiles.xml can't be found/loaded from the default location (though of course its impossible to tell without a full debug log), and so not a skin shortcuts error...

Nessus85100:
- `widgetName` is fixed
-  multiple instances of management dialog, partially fixed. Now no matter how fast you click is not opened more than 2-3 times. But still can be open more than once for sure.

Cheers
Nessus

ghost:
Thanks for the feedback. Not sure much more can be done about #40 as I suspect the multiple clicks are coming before the script has been loaded the first time (aside from some refactoring of the scripts entry point - default.py is recompiled each time the script is run, so it's possible to move all its code into a secondary python file which wouldn't be recompiled each run, and create a minimal default.py which would speed up script launch by a few fractions of a second). But feel free to re-open that issue if you wish :)

marcelveldt:
@Nessus85100 I fixed some small issues in the skinshortcuts script and added in a few extra features (see documentation), for example how to override actions. Do you mind grabbing the latest version from Git to test ?

Also, I have released a first beta for a skinhelper service. Don't know if you're interested in that, but have a look if you want:

http://forum.kodi.tv/showthread.php?tid=235676

Nessus85100:
I am kinda busy this week but as soon as i find some free time will do it.

Thanks
Nessus

PR #43, opened 2015-08-12
fixed issue when selecting playlist the result is always action play

marcelveldt:
There was a typo in the code where the user selects a playlist and answers the question to display or play the playlist. Because of that, the result would be that the playlist will always be played instead of the user preference.


ghost:
Very nice spot :)

PR #45, opened 2015-08-14
always use pretty dialog for selecting backgrounds

marcelveldt:
allows any $INFO backgrounds to be shown too.
with this change it has the same behaviour as the select icon dialog

### DIFF
--- a/resources/lib/gui.py
+++ b/resources/lib/gui.py
@@ -730,6 +730,9 @@ class GUI( xbmcgui.WindowXMLDialog ):
         if tree is not None:
             elems = tree.findall('thumbnail')
             for elem in elems:
+                if "condition" in elem.attrib:
+                    if not xbmc.getCondVisibility( elem.attrib.get( "condition" ) ):
+                        continue
                 self.thumbnails.append( [elem.text, DATA.local( elem.attrib.get( 'label' ) )[2] ] )
                 self.thumbnailsPretty[elem.text] = DATA.local( elem.attrib.get( 'label' ) )[2]
             
@@ -1203,18 +1206,23 @@ class GUI( xbmcgui.WindowXMLDialog ):
                         backgroundLabel.append( key[1].replace( "::PLAYLIST::", playlist[1] ) )
                         backgroundPretty.append( LIBRARY._create(["", key[1].replace( "::PLAYLIST::", playlist[1] ), "", {}] ) )
                 else:
-                    background.append( key[0] )     
+                    background.append( key[0] )
+                    virtualImage = None
+                    if key[0].startswith("$INFO") or key[0].startswith("$VAR"):
+                        virtualImage = key[0].replace("$INFO[","").replace("$VAR[","").replace("]","")
+                        virtualImage = xbmc.getInfoLabel(virtualImage)
+
                     log( repr( defaultBackground ) + " : " + repr( key[ 0 ] ) )
                     if defaultBackground == key[ 0 ]:
                         backgroundLabel.append( key[1] + " (%s)" %( __language__(32050) ) )
-                        if xbmc.skinHasImage( key[ 0 ] ) == True:
+                        if xbmc.skinHasImage( key[ 0 ] ) or virtualImage:
                             usePrettyDialog = True
                             backgroundPretty.append( LIBRARY._create(["", key[ 1 ] + " (%s)" %( __language__(32050) ), "", { "icon": "DefaultFile.png", "thumb": key[ 0 ] } ] ) )
                         else:
                             backgroundPretty.append( LIBRARY._create(["", key[ 1 ] + " (%s)" %( __language__(32050) ), "", {} ] ) )
                     else:
                         backgroundLabel.append( key[1] )
-                        if xbmc.skinHasImage( key[ 0 ] ) == True:
+                        if xbmc.skinHasImage( key[ 0 ] ) or virtualImage:
                             usePrettyDialog = True
                             backgroundPretty.append( LIBRARY._create(["", key[ 1 ], "", { "icon": "DefaultFile.png", "thumb": key[ 0 ] } ] ) )
                         else:
@@ -1286,7 +1294,7 @@ class GUI( xbmcgui.WindowXMLDialog ):
             # Generate list of thumbnails for the dialog
             for key in self.thumbnails:
                 log( repr( key[ 0 ] ) + " " + repr( key[ 1 ] ) )
-                thumbnail.append( key[0] )            
+                thumbnail.append( key[0] )
                 thumbnailLabel.append( LIBRARY._create(["", key[ 1 ], "", {"icon": key[ 0 ] }] ) )
             
             # Show the dialog


ghost:
-1 from me as-is

The reason that the background select function has two possible dialogs (pretty and ugly) is because there are two different ways to use Skin Shortcuts to manage backgrounds

(1) As I presume you are, using the script to define a number of specific images in the overrides `<background />`, and using the dialog to pick a particular image. Here, it makes sense to use the pretty dialog so the background can be previewed (and, yes, makes sense to allow it to display `$INFO` based backgrounds too).

(2) As I do, the `<background />` overrides define a background _style_ - fanart slideshow, ken burns, disc cases, tellytubbies (one for the grandkids!) - not specific images. As such there is no image to display in the select dialog, so no point using the pretty one. Moreso, if the pretty one is used all the items will have exactly the same image (defaults to DefaultFile.png), which is the wrong image when a file isn't being selected. So, to my mind, it's better at present to use the ugly dialog when the background functions are being used in this way :)

A few possibilities:
- Have the pretty dialog by default, but add an override so the ugly can be used if desired (and update docs!)
- Have only the pretty dialog, but extend the `<background />` override to include a preview attribute, so when using background styles, there can be a specific image for each background (and update docs!) - actually I quite like this idea...
- Extend the checks of whether of not to use the pretty dialog to take into account whether the `[backgroundID]` is an `$INFO`
- Overrule me. I can get used to a pretty dialog, and can add my own workaround if it really annoys me :)

Other thing to note, if you do decide to go with pretty dialog all the time, then you can get rid of the `usePrettyDialog` code.

marcelveldt:
aha! Now I know where that difference between "pretty" and none pretty comes from :-)
Thanks for the explaination. Will revert this PR (and come up with something better)..

In that case...

I have a a whole bunch of backgrounds specified in the overrides.xml
Both dynamic ones (with $INFO or $VAR) and static images.
Before this change it only displayed the static images in the preview and the INFO ones were blank...
Any clue how to fix that with leaving the pretty/non pretty selection in there ?

Also, another challenge (yeah I'm heaviliy modifying my skin to work with latest skinshortcuts version)... I have a whole bunch of custom shortcuts specified. Now I'd like to use the backgrounddefault override to set default background for those shortcuts so when someone adds it to the homescreen, the background will be set correctly by default.

While working at this I have a few questions:
1. Is it possible to use both icon and thumb for a skin provided shortcut ? That way you can have both a dynamic INFO image as thumb and the icon as fallback...
2. Is it possible to directly provide background= as property on the shortcuts? That saves a lot of code with backgrounddefault stuff

btw: I seem to have discovered an issue with the backgrounddefault. I can set it perfectly fine for default/common items but for my dynamic shortcuts it doesn't work. Also the documentation explains that you can use both labelID and defaultID for the backgrounddefault but only labelID seems to be working:

this is working fine:

```
<backgrounddefault labelID="livetv">special://skin/extras/backgrounds/hover_my tv.jpg</backgrounddefault>
```

and this is not:

```
<backgrounddefault defaultID="livetv">special://skin/extras/backgrounds/hover_my tv.jpg</backgrounddefault>
```

and this isn't working either:

```
<backgrounddefault labelID="info-window-home-property-emby-nodes-0-title">$INFO[Window(Home).Property(AllMoviesBackground)]</backgrounddefault>
```

Any clues ?

ghost:
> I have a a whole bunch of backgrounds specified in the overrides.xml
> Both dynamic ones (with $INFO or $VAR) and static images.
> Before this change it only displayed the static images in the preview and the INFO ones were blank...
> Any clue how to fix that with leaving the pretty/non pretty selection in there ?

Check if the key[ 0 ] starts with $INFO or $VAR, and ask Kodi to resolve it for you? TBH, however you want to do it :)

> 1. Is it possible to use both icon and thumb for a skin provided shortcut ? That way you can have both a dynamic INFO image as thumb and the icon as fallback...
> 2. Is it possible to directly provide background= as property on the shortcuts? That saves a lot of code with backgrounddefault stuff
1. Sorry, nope. Decision made a looong time ago whilst ReFocus & the script were being modified to work with each other, to (generally) allow skinner only to edit the icon, and let the user only edit the thumb - so that the users choice is always the final choice. It's kinda hardcoded now, and difficult to work around it.
2. Again, sorry no. As the widgets and backgrounds and other properties are skin specific (and because ultimately the code to load them is the same code whether we're loading the skin defaults or the users custom shortcuts) the properties and the shortcuts are kept separately.

> I seem to have discovered an issue with the backgrounddefault. I can set it perfectly fine for default/common items but for my dynamic shortcuts it doesn't work. Also the documentation explains that you can use both labelID and defaultID for the backgrounddefault but only labelID seems to be working:

I'll try and take a look into this tonight. If I don't manage it'll be next weekend before I can get to it (I'm away with the family this coming week, and it's a very strict no-internet/coding week away :sob:)

marcelveldt:
thanks for your answers!. In the meanwhile I managed to fix the visibility condition on icons, that was something on the TODO list too.

I will try to look into the background dialog issue. Step by step I'm getting my way through the code ;-)

Enjoy your family trip! Yeah, sounds familiar about that no-internet/coding :-)

ghost:
As regards the background defaults, I've had a visual look at the code and spotted one issues - line 531 of datafunctions.py - `"mainmenu"` should read `elem.attrib.get( "group" )`. Don't think that would affect the issue you're seeing, though.

marcelveldt:
problem with the Original PR is resolved. Both the pretty and the non pretty dialog are now working again and I added in a small check for the INFO labels.

Also fixed is the visbility condition on the thumbnails.

ghost:
I know I originally suggested it, but any need for `xbmc.getInfoLabel`? virtualImage isn't actually passed to the `LIBRARY._create` function, so could it just be set True/False based on whether or not the key starts with $INFO/$VAR...? Other than that, this looks good to go :)

marcelveldt:
I've added that so it will also check if the image actually exists. Maybe it's a bit overkill but couldn't harm ?

ghost:
I'm sure I'm just missing it, but where...?

`if xbmc.skinHasImage( key[ 0 ] ) or virtualImage:` is checking for the key[ 0 ] image, but is only checking whether virtualImage is an object (or whatever the literal Python translation of that is).

`if xbmc.skinHasImage( key[ 0 ] ) or xbmc.skinHasImage( virtualImage ):` what you meant to do...? :)

marcelveldt:
virtualImage will return False if the getInfoLabel returned None or ""
It's the pythonic way ;-)

ghost:
But getInfoLabel isn't specifically checking for an image, is it?

marcelveldt:
yes it is... virtualImage = xbmc.getInfoLabel(virtualImage)

ghost:
Only if the $INFO/$VAR is an image URI, though ;)

marcelveldt:
yep. .. so either it's no VAR. then it will Always be None and if the infolabel is retrieved and returned empty it's also None...

:-)

ghost:
And my head is starting to hurt ;) I think it's fair enough as is - just something to be aware of, trying to use $INFO/$VAR's when using `<background />'s` to manage the background style will trigger the pretty dialog. But that's no huge issue, so let's get this merged :)

marcelveldt:
thanks !

I will go ahead and try to get that defaultbackground thing fixed

marcelveldt:
It's a strange problem with that backgrounddefault. If I fill in the group it's working a bit better.
In that case the dialog in the background selection displays (default) after the title but still it isn't built into the menu.

ghost:
D'you mind fixing a security issue for me whilst you're doing that? (I've already fixed it for .hashes, but overlooked it for properties!)

datafunctions.py:- 

`import ast`

line 462 (in _get_additionalproperties)

from `listProperties = eval( file )`
to `listProperties = ast.literal_eval( file )`

This prevents arbitrary python code being run from a modified .properties file. (The eval method is also used for upgrading existing shortcuts further down in the py file, but as those methods are earmarked for removal, it's probably safe to ignore them for the moment).

Much appreciated :)

ghost:
Did you see the bug I spotted when `group` is defined...? (See further up in the conversation...)

marcelveldt:
yep, seen that bug, fixed it just now but doesn't help.

I'll fix the security issue

ghost:
From what I can see from looking through the code, the defaults are being loaded correctly into the defaultProperties. That brings the issue down to one of three areas
1. Particularly for the third example you spotted, the `<backgroundDefaults />` are being incorrectly read from the overrides.xml file - should be easy enough to add some logging to check that the values are being set as you expect them in getAdditionalProperties() (`log(repr(defaultProperties))`).
2. The properties are being loaded correctly, but aren't being assigned to the right menu items. That all happens in datafunctions.py checkAdditionalProperties() - line 753. Again, should be easy enough to check whether the right properties are being assigned.
3. Or the properties aren't being written to the menu items as they should be. That all happens in xmlfunctions.py - from around line 710.

marcelveldt:
thanks for the info, that helps in debugging the info.
So far I added some logging to find out that the labelID and defaultID return None in datafunctions.py line 481. I'll dig some more to see why

ghost:
If it's returning None for both then that's an issue, assuming any of the default elements (`widgetdefault`, `widgetnoddefault`, etc) have an labelID or defaultID attribute. Case mismatch, possibly...?

marcelveldt:
I tried both labelID and defaultID and with and without specifying the grouep:

```
<backgrounddefault group="mainmenu" labelID="movies">$INFO[Window(Home).Property(AllMoviesBackground)]</backgrounddefault>
    <backgrounddefault group="mainmenu" defaultID="tvshows">$INFO[Window(Home).Property(AllTvShowsBackground)]</backgrounddefault>
```

ghost:
Neither worked?

marcelveldt:
Nope, both examples fail...
I've added some print lines and both defaultID and labelID are empty

ghost:
Working fine for me (having added those lines to my Conq forks overrides.xml, as that skin is set up on my development machine at the moment.) I added the following debug lines to _get_additionalproperties, for both `elemSearch[ 0 ] == "background"` (so for both with and without `group` specified):-

```
log( "default type: %s" %( elemSearch[ 0 ] ) )
log( "labelID: %s" %( labelID ) )
log( "defaultID: %s"  %( defaultID ) )
```

And got the following in my debug log:-

```
19:48:50 T:123145302839296   DEBUG: script.skinshortcuts: default type: background
19:48:50 T:123145302839296   DEBUG: script.skinshortcuts: labelID: movies
19:48:50 T:123145302839296   DEBUG: script.skinshortcuts: defaultID: movies
19:48:50 T:123145302839296   DEBUG: script.skinshortcuts: default type: background
19:48:50 T:123145302839296   DEBUG: script.skinshortcuts: labelID: None
19:48:50 T:123145302839296   DEBUG: script.skinshortcuts: defaultID: tvshows
```

which is exactly what I would expect.

marcelveldt:
strange... those loglines are exactly what I added. I suspect that there is some error in my overrides file or something. I'll strip it down to a minimum to see what it does.

ghost:
If I can help, feel free to e-mail me a copy of your overrides - eeeeepish@gmail.com (this is my spam e-addy, so let me know before you send anything to it so I know to actually check it!) - and I'll see if I can reproduce it with what you're working with.

marcelveldt:
OK, I managed to fix it again. The labelID and defaultID is no longer blank.
It was a small typo in some code I changed while debugging the main problem...
So... This issue was caused by myself  :-)

So, now I'm back to where I started:

The backgrounds dialog "detects" the default because it's showing the (Default) prefix.
It is however not built into the xml.

ghost:
Glad you fixed it, because I was driving myself mad trying to work out what situation would cause those values not to be set, and I really couldn't see where that could happen (and I've just deleted a seven-paragraph response detailing why what you were saying wasn't possible!)

So we come back to the original three areas I specified above.
1. The defaults aren't being loaded correctly (getAdditionalProperties)
2. The defaults aren't being set to the correct menu item (checkAdditionalProperties)
3. The menu item isn't being written correctly (xmlfunctions.py, line 710 onwards)

:)

marcelveldt:
hehe, I found it strange myself too because it was going worse :-)
Just a little note: When I added some logging I noticed that the same lines come by many times. Is that additionalproperties method called multiple times ?
This is just an FYI: there might be a small speed improvement there :-)

I will check the above mentioned 3 levels step by step to figure it out.
Thanks!

ghost:
I noticed the same thing and, yes, there's a potential speed improvement (though at this point we're talking fractions of a second!) to be gained there :)

marcelveldt:
Yeah, I know it will only be some milliseconds but I thought to let you know.
BTW: I'm at 1200+ lines in my overrides.xml and the script is still blazing fast. So that latest rework of you to improve the speed really helped.

ghost:
I know it sounds crazy, but I'm almost a little annoyed at how fast it is these days, just because on any kind of modern system (ignoring systems like the rPi, where my current build takes approx 10 seconds) I had to split the build process into two parts - build menu and write menu - otherwise the first part was so fast it looked like the second had hung (for a second at least!) :P

Edit twelve _thousand_ lines? _twelve_ thousand lines??? Might be worth us having a conversation at some point about what changes can be made to the script to reduce that somewhat :) (The original point of the overrides.xml was just to override actions, and it sounds like it's become a little out of control!)

Edit 2: It's the night before my holiday, so yes, I misread that and I blame alcohol. But still twelve _hundred_ lines? _Twelve_ hundred lines? ;)

marcelveldt:
:-)

marcelveldt:
OK, here's what I have found so far:

1) backgrounddefault only works if you provide the labelID while it should also work with defaultID.
I've been looking through the code but i can't seem to find where the background is checked with the labelID or defaultID.

2) If you select one of the skin provided shortcuts to add to your menu, the background isn't set while it should be set to the default background. It does work correctly (except the defaultID issue) if the shortcut is provided in the skin provided mainmenu.DATA.xml. If you select a new shortcut to your menu menu things like the label and icon are set correctly but the background isn't.

About my overrides.xml:  Well the main part it is so big is because of the "smart shortcuts" feature, besides that I provide lot's of stuff for the user to choose from. Maybe some parts can be migrated to skinshortcuts some day ;-)

ghost:
1. Assigning to a menu item based on its labelID / defaultID happens in datafunctions.py checkAdditionalProperties. If this is failing, this is _where_ its failing.
2. You mean creating / setting a new shortcut? Without checking (and I'm pretty much out of time to do detailed checking), I'm not sure there's any code in place to do that.

marcelveldt:
1) ok thanks, I will look into it. At the end of the day I have learnt quite a lot more about the script ;-)
2) Ah! So you mean this was never implemented ? Well, in that case I'll try to find out how to add it

ghost:
1. Good luck :)
2. Probably (there's things I need to learn about the script myself - and, to be perfectly honest, its become such a complex script that I'm not 100% sure what the interactions are in certain circumstance). To the best of my knowledge, there isn't any code that will run to assign properties (be them widgets, backgrounds, etc) to a new shortcut that is added.

marcelveldt:
Ah, okay. That explains it. Well, if you don't mind I'll try to add that because it's a little weird that the user doesn't get a good default setting...
In my opinion it should be as easy as selecting a new shortcut and (if provided by the skin) that shortcut will have it's widget and background setup immediately.

I had all this code previously stored in skin settings and now I'm finally migrating it all to skinshortcuts (thanks to the speed changes) so that's why I never noticed it.

ghost:
Absolutely agree - a new shortcut should have all the skin-provided properties pre-set. If that's not currently happening (as I say, this is getting into a level of script-complexity I haven't looked at in far too long), than any PR that fixes that is most welcome :)

(To be clear, when resetting shortcuts to skin-defined, all properties _should_ be correctly set and, if they're not, that's another issue).

Here's hoping that your upcoming PR fixes for all additional properties - widgets, backgrounds, and anything else the skin sets! Looking forward to it if, for no other reason, it greatly simplifies the situations this needed to be fixed for (if it works correctly, this should come down to two - fixed - situations, default shortcuts and added shortcuts),

marcelveldt:
ok, the default widget and background is done. 
only problem left is the defaultID for defaultbackground upon skin xml build. I'll give it one more try to fix it, else I'll create a bug report so we can pick it up later.

ghost:
Do feel free to create a bug report, which I'll look at when I return (I'm about half-an-hour from heading to bed, and still have some packing to do). Just to note, because I almost put this as a comment on one of your previous commits until I read it properly, skins cannot specify a defaultID, only a labelID (which then becomes the defaultID when the shortcut it loaded/changed - similar to how skins can only provide icons, not thumbs). Not sure its causing your issue with the default background, but it worth bearing in mind as you appeared to be getting it wrong in an earlier comment/commit :)

If not, I look forward to seeing what you've come up with when I get back off holiday :)

ghost:
To be clear, I'd be happy to merge the upcoming commits myself if:-
1. When adding a new shortcut, its properties are set to the skin-provided properties.
2. When resetting all shortcuts, shortcuts are reset to the skin-provided properties.

Look forward to seeing these features when I return from holiday :)

marcelveldt:
both 1 and 2 are now working as intended. I've also managed to finally squash that defaultID problem.
It was a None type check wrongly formatted. There was: if DefaultID == None while that should be if not DefaultID

anyway, it's fixed now. :-)

marcelveldt:
enjoy your holiday!

PR #46, opened 2015-08-14
Only evaluate default overrides once

ghost:
Another build speed improvement, albeit with a fraction of a second improvement - only load shortcut properties and defaults once, rather than repeatedly for each menu item. Also two other fixes:-
- Security fix - arbitrary Python code could be run from a compromised .properties file
- Correctly set group for `backgrounddefault` overrides with group attribute set.


ghost:
*keeps fingers crossed that testing after it's merged, whilst working out that defaultbackground problem, doesn't find issues I missed ;)

Cheers. :)

marcelveldt:
It's working fine... just tested :-)

ghost:
Good to know :) Thanks again.

(And apologies for where my sense of humour goes when I'm drinking!)

PR #47, opened 2015-08-14
assign default background and widget when selecting a new shortcut

marcelveldt:
@Ignoble61  as discussed... This sets the default background and/or widget when selecting a new shortcut.

### DIFF
--- a/resources/lib/gui.py
+++ b/resources/lib/gui.py
@@ -745,7 +745,6 @@ class GUI( xbmcgui.WindowXMLDialog ):
     # === GUI INTERACTIONS ===
     # ========================
 
-    
     def onClick(self, controlID):
         if controlID == 102:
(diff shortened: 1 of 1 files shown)


ghost:
I apologise that I don't have time to test this properly, so I'll ask just one question which, based on our previous conversation, will confirm that this works as intended (assuming that you've actually tested it in practice...!)

Question: You have tested that, without this code, adding a new shortcut doesn't load default properties provided by this skin, and -  with it - the new shortcut does have these properties?

Assuming your answer to this is 'Yes', as I have every reason to believe it works perfectly and I will happily merge this :)

marcelveldt:
Yes, I've tested it and it does exactly that :-)

ghost:
I presumed you would have, but it seemed sensible to just confirm it :) Very happy to merge this.

PR #50, opened 2015-08-19
Remove xpath search for Python 2.6 compatibility

### LINKED ISSUE #49: Script broken on Python 2.6
Sorry, I don't have time to fix this myself at the moment (not even meant to be using the internet, but modern phones are a wonderful thing!), but a note on recent changes. Using xpath with elementtree is not supported with Python version 2.6, so using it is breaking the script on systems with that Python version (see logs on fTV skin thread, for example)

E.g. `if not newelement.findall(".//property[@name='path']"):` will cause a script crash on systems using that Python version.

ghost:
Closes #49, and fixes menu build on systems with Python 2.6 including Mac and FTV. Happy with the fix, @marcelveldt? :)

### DIFF
--- a/resources/lib/xmlfunctions.py
+++ b/resources/lib/xmlfunctions.py
@@ -692,14 +692,14 @@ class XMLFunctions():
                 onclickelement.text = onclick.text
                 
             # Also add it as a path property
-            if not newelement.findall(".//property[@name='path']"):
+            if not self.propertyExists( "path", newelement ):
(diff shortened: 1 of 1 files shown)


marcelveldt:
Very happy, thanks for the fix !

And uhm... You shouldn't be coding now ;-) Enjoy your holiday!

ghost:
My real holiday is next week, when I'm back home and the grandkids have been dispatched ;)

Issue #52, opened 2015-08-20
Widgets select


Nessus85100:
I've tested the latest commit and everything seems to working fine except one thing. I use this
`<label2>$INFO[Container(211).ListItem.Property(widgetName)]</label2>` in 312 button to show the name of the current selected widget. After the latest commits (i don't know when but it was after 0.5.1 BETA) the widget name has a $INFO prefix in front.
Example...
Widget name: Popular Movies
now it shows: $SKIN[31251]|skin.bello|Popular Movies (31251 is my skin string for "Popular Movies")

This is happening only with the skin strings. Kodi and script's strings are working fine. I remember i've read somewhere in the documentation that skin strings should have a ::LOCAL:: prefix but until now they were working fine without that.

I have also some requests if i may. We really need to find a way to determine the kind of "Sources" widgets. I mean when user selects a Source as a widget there is no way that the skin to know what kind of source is that (movies, tvshows, music etc) so to display the proper thumbs and info. Is it possible to add some question to the user about the type of the source content and then set that as `widgetType` ?. That will make things more easy to handle. If there is no way to determine (and add a property about that) the source type then i will remove them from my widgets list since there is no point to have them and not to be able to display them properly.

One last thing. Is possible to break the the library nodes in additional nodes ?.
What i mean is this... This will show all the Video Library nodes:

```
<node label="32030">
   <content>video</content>
</node>
```

Now, is it possible to break this in additional nodes like this ?:

```
<node label="32030">
   <content>movie</content>
   <content>tvshow</content>
   <content>files</content>
.
.
</node>
```

I've try the above code but is not working. This could help to group very well the library nodes widgets and the skin widgets.

EDIT: I forgot about this and i have no idea if is old but i just notice it. If you try to override the PVR Guide icon with the Kodi 22020 string you will end up with the DefaultAddon icon...
`<icon labelID="22020">special://skin/extras/icons/Guide.png</icon>`

Cheers
Nessus

marcelveldt:
How did you specify your widget's name in your overrides.xml ? It is enough to just set the numeric ID of the language string. No need to add $LOCALIZE[], so for example:

``` xml
<widget label="31196" condition="true">popular movies</widget>
```

should suffice.

[...]

PR #53, opened 2015-08-21
Widget fixes

ghost:
As reported in #52 
- Fix unicode error in select shortcut dialog when generating widgetName
- Fix skin-localised strings displaying in management dialog

### DIFF
--- a/resources/lib/library.py
+++ b/resources/lib/library.py
@@ -7,6 +7,7 @@ from xml.dom.minidom import parse
 from xml.sax.saxutils import escape as escapeXML
 from traceback import print_exc
 from unidecode import unidecode
+from unicodeutils import try_decode
 import datafunctions, nodefunctions
 DATA = datafunctions.DataFunctions()
 NODE = nodefunctions.NodeFunctions()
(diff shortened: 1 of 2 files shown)


resources/lib/library.py
```diff
@@ -1500,9 +1501,9 @@ def explorer( self, history, location, label, thumbnail, itemType, isWidget = Fa
         widgetType = None
         addonType = None
 
-        dialogLabel = label[0].replace( "  >", "" )
+        dialogLabel = try_decode( label[0] ).replace( "  >", "" )
         if len( label ) != 1:
-            dialogLabel = label[0].replace( "  >", "" ) + " - " + label[ len( label ) - 1 ].replace( "  >", "" )
+            dialogLabel = try_decode( label[0] ).replace( "  >", "" ) + " - " + try_decode( label[ len( label ) - 1 ] ).replace( "  >", "" )
```
phil65, line 1507: pythonic way for "label[ len( label ) - 1 ]" is "label[:-1]" (you want to remove the last char, right?)

ghost, line 1507: `label` is actually an array of strings, so it's not removing the last char, it's getting the last element of the array. Probably still not doing it the pythonic way, though ;)

phil65, line 1507: in that case "label[-1]" ;)

ghost, line 1507: Y'know, I'm starting to hate Python every time I return to it ;) Cheers Phil I'll update it right away.

marcelveldt:
Thanks for fixing!

PR #54, opened 2015-08-25
If no label, redirect GUI 307 to 401

ghost:
When changing action in management dialog (GUI 307), if the shortcut being changed doesn't have a label the script will now redirect to select shortcut (GUI 401).

As the script will only save user shortcuts with a label (the default label of `<None>` is used to decide whether to save a shortcut), if the user selects 'Change action' on a new shortcut their selected shortcut won't be saved unless they also manually edit the label. By redirecting to 401 in this instance, the label will be set when they select their action and so the shortcut will be saved.

@marcelveldt can you confirm this doesn't interfere with what your aims are for the updated change action function :) Cheers.

### DIFF
--- a/resources/lib/gui.py
+++ b/resources/lib/gui.py
@@ -997,6 +997,11 @@ class GUI( xbmcgui.WindowXMLDialog ):
             log( "Change action (307)" )
             listControl = self.getControl( 211 )
             listitem = listControl.getSelectedItem()
+
+            # If no label, redirect click to 401 (Select Shortcut)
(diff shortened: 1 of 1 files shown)


ghost:
Can I assume no-one has any issues with this being merged? :)

marcelveldt:
no problems here ;-)

ghost:
Good thing, I was literally about to merge it myself ;)

marcelveldt:
hehe, I didn't notice the PR sooner

Issue #58, opened 2015-09-01
ready for deploy to stable ?


marcelveldt:
The current beta version has been tested by a bunch of users. Some issues found and fixed.
Now things seem to be very stable and this new version offers a lot of improvements.

Everybody happy with a deploy to the Kodi repo ?

ghost:
It's probably worth considering whether any of the fixes requested in #57 can go in before a push to stable - I'm not particularly familiar with the code in question, but some of the requests there (I'll go into detail in the PR itself) seem pretty straightforward and unlikely to causes breakages. Good to get them fixed before Tomer has to wait for another round of updates/beta's (however long that may take) before they get onto the repo?

For me, other than that, it's good to go (aside from removing 'Beta' notices from the changelog/readme). @Nessus85100 would be good to get approval from too :)

marcelveldt:
I already told Tomer that we're having a little feature freeze first to update the stable release before adding in any new features but if you're confident that that change won't break anything and you have time to look into it no problem to wait a few more days. It's just that the current beta has actually been tested by users. If we add something we should publish a new beta for a few days first.

Totally agree with getting the feedback from @Nessus85100 because he has tested the widget changes very well.

ghost:
Not something I'll look into soon as I'm not overly familiar with that code and I'm working primarily on other projects at the moment. (Actually not sure offhand, is it your code initially...?) I'm happy to wait until I or anyone else gets to it :)

marcelveldt:
Yeah, I've worked on that code a few weeks back so I'm familiar with it. I'll pick it up once the current release has been deployed on the Kodi repo.

ghost:
Just remembered one more thing that it's worth considering before a stable push - do you, @marcelveldt, want to consider adding the skin.helper widgets as default into the scripts overrides.xml widget-grouping? (With visibility condition to ensure they only display if the script is installed, of course...)

Nessus85100:
Yep, all working fine for me. I think we are good to go.

By the way, excellent work guys !!!

Cheers
Nessus

BigNoid:
Indeed, great work!  
I don't use many features of the script, but all is working fine in Nox.

ghost:
> Yep, all working fine for me. I think we are good to go.
> 
> Indeed, great work!
> 
> I don't use many features of the script, but all is working fine in Nox.

[...]

PR #59, opened 2015-09-04
Mark stable

ghost:
Mark script.skinshortcuts as stable in all documentation, as well as updating the version number so all using beta versions are using stable version (0.5.3).

(Stable release based on discussion in #58.)

Whilst I'm happy to contribute to a stable release, I'm not comfortable actually requesting a repo-push (actually, I personally have no real interest in the repo version as opposed to the version that I make available to my home and my family), so I push this in the presumption that there others - which has actually become a decent number of individuals involved in the script - who are happy to take care of any necessary push to repo.

It also makes two further assumptions - the script should still be compatible with Frodo, the initial version the script was created and is still compatible with - and the assumption is that Team Kodi are still accepting updates for scripts on the Frodo repo - if not, there should be changes to the default shortcuts provided by the script, to Library.py to remove reference to the Frodo-era PVR shortcuts and to the xbmc.python version  in the addon.xml. And that @marcelveldt doesn't want to add default support for skin.helper widgets (or wants to add them, and then push) to the current script, as mentioned again in #58 :)

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -1,7 +1,7 @@
 script.skinshortcuts was written with the intention of making user customizable shortcuts on the home page easier for skinners.
 
 
-What's New for Skinners (version 0.5.2 - beta release)
+What's New for Skinners (version 0.5.3)
 -----------------------
 
  - Add additional actions on a per-menu basis - see "Advanced Usage.txt", section "Overrides.xml", part 16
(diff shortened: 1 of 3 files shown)


marcelveldt:
Great work!
I will request the repo update tomorrow.
That will be for Helix + Isengard. Gotham + Frodo are no longer supported.

ghost:
If it's for Helix, and if my memory of when the PVR shortcuts changed, then - as said above, it's worth removing references to the Frodo-era PVR shortcuts. If my memory of when these changes is correct, let me know and I'll make a further PR removing all Frodo-era references :)

(Edit - Gotham (- I genuinely forget about this version!) = XMBC/Kodi 13, right...?

marcelveldt:
yeah Gotham = XBMC 13 :-)

PR #60, opened 2015-09-04
Remove references to Kodi v13

ghost:
Makes minimal version for script v14 - previous Kodi versions will be missing available shortcuts, both in available shortcuts and built menus. Assuming minimum version for repo push is now v14.

Assumes conversation regarding minimum Kodi version in #59 is correct :)


marcelveldt:
looks good. another bit of cleanup.
The new version 0.5.3 will be pushed to the Helix+isengard repo. So indeed no longer to Gotham.
Offcourse the previous version of the script will stay on the Gotham repo untill it's eol.

PR #61, opened 2015-09-04
skinhelper widgets + fixed scripts not visible in program addons listing

marcelveldt:
1) add skinhelper script widgets by default to the overrides.xml
2) fixed small issue that program-addons was not showing scripts as executables in the listing

### DIFF
--- a/resources/lib/library.py
+++ b/resources/lib/library.py
@@ -1322,7 +1322,6 @@ class LibraryFunctions():
                         if item['enabled'] == True:                            
                             path = "RunAddOn(" + item['addonid'].encode('utf-8') + ")"
                             action = None
-
                             thumb = "DefaultAddon.png"
                             if item['thumbnail'] != "":
                                 thumb = item[ 'thumbnail' ]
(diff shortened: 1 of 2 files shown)


marcelveldt:
@Ignoble61 do you mind take a look at this ?

ghost:
I was expecting to see favourites added to this, and had prepared comments about how I had purposefully left those out of the new widget additions as they can't be relied upon to provide useable widgets ;) As is, this is a great addition and I'm very happy to merge this!

Edit: I also want to congratulate you on the integration of your skin.helper script - though I wrote the code that you've used to enabled third-party script integration via "::BROWSE::", this isn't something I'd ever even remotely considered, and it's fabulous to see that you've been able to exploit this to allow 3rd party script-integration. I've long avoided providing specific 3rd party integration (most notably cinema experience and advanced launcher) into the script because I expected it to require significant core code changes. The fact that you've been able to do it so simply (relatively, and primarily - though I'd ideally like to see this done exclusively - externally) is, quite frankly, fantastic.

PR #63, opened 2015-09-05
Fix - Just Select methods not parsed to python booleans

ghost:
As raised by JurialMonkey on Kodi forums.

### DIFF
--- a/default.py
+++ b/default.py
@@ -176,6 +176,14 @@ class Main:
         self.LIST = params.get( "skinList", None )
         self.CUSTOM = params.get( "custom", "False" )
         self.NONE = params.get( "showNone", "False" )
+        if self.CUSTOM == "True" or self.CUSTOM == "true":
+            self.CUSTOM = True
+        else:
+            self.CUSTOM = False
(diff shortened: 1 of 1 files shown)


default.py
```diff
@@ -176,6 +176,14 @@ def _parse_argv( self ):
         self.LIST = params.get( "skinList", None )
         self.CUSTOM = params.get( "custom", "False" )
         self.NONE = params.get( "showNone", "False" )
+        if self.CUSTOM == "True" or self.CUSTOM == "true":
```
phil65, line 179: As an idea, self.CUSTOM.lower()  == "true"

ghost:
Yup, that's my bad. I actually avoided using .lower() on purpose, thinking that it could possibly fail if self.CUSTOM wasn't a string (that and my whole OCD about always trying to check that something is what I require and expect before I do anything to it :wink:). It's only after it got merged I realised that only strings can be passed into the script and, if no value was passed in, the line right above is making sure it's a string rather than None. Definitely a change worth making :)

PR #64, opened 2015-09-11
improve plugin content detection

marcelveldt:


resources/lib/library.py
```diff
@@ -1383,28 +1383,32 @@ def detectPluginContent(self, item):
...
-            if item["artist"][0] == item["title"]:
+            if item["type"] == "artist" or item["artist"][0] == item["title"]:
                 return "artists"
-            elif item["album"] == item["title"]:
+            elif item["type"] == "album" or item["album"] == item["title"]:
                 return "albums"
-            elif (item["type"] == "song" or (item["artist"] and item["album"])):
-                return "songs"
+            elif (item["type"] == "song" and not "play_album" in item["file"]) or (item["artist"] and item["album"]):
+                return ("songs", image)
```
ghost: I'm sure this is correct, but I have to question it (just looking at this code, not the context) - every other instance returns just a string, but here you're returning a tuple?

ghost:
Something to consider for the next version of Skin Shortcuts - and something very much up for debate :) :-

It occurs to me that some of the widget functionality we're now using your amazing skin.helper script for could be done directly in Skin Shortcuts via `<shortcut />` elements - the library data provider widgets, for example. This would remove the need for their content to be detected, reduces the reliance on external scripts, and would mean they'll work with the `<widgetdefaultshortcut />` that it looks like it's going to be sensible to introduce in a new version of the script.

Just to be clear, I'm not suggesting removing the skin helper elements in the groupings, but looking at moving what can be done directly by this script. Things like the smart shortcuts, for example, have to stay :)

marcelveldt:
Sounds good to me. I had that same idea but didn't want to "overwhelm" the skinshortcuts script ;-)

resources/lib/library.py
marcelveldt: oops, that's an error ;-)

ghost:
It definitely makes sense not overwhelming things, particularly in a 1.0 as-it-were of the new widgets. There's been a few issues with usability (from a skinners perspective) with them, so I'm starting to broadly plan the necessary updates - I'll include switching to `<shortcut />'s` where appropriate :)

ghost:
After the typo fix it looks good to me, and I'm assuming you've tested it :) If you can merge your commits, I'll merge the PR :)

marcelveldt:
I've squashed the commits.

BTW: I created a tag for the update to the Kodi repo so it's safe to modify the code.
The 0.5.3 tag will be pushed to the Kodi repo.

ghost:
Great to know about the tag. I'm working on another project at the moment, but it's good to know I can start modifying when I feel like it ;)

PR #66, opened 2015-09-11
Support adding music nodes from library node editor

ghost:
As the title says, extends the current code to allow music nodes to be added to the main menu (when using a skin that uses the script to manage the whole menu) from the context menu of the Library Node Editor (reliant on a PR about to be requested for that script) - some work may be needed for this completely overlooked feature to be updated in both scripts together!

Library node editor PR - https://github.com/XBMC-Addons/plugin.library.node.editor/pull/1

### DIFF
--- a/default.py
+++ b/default.py
@@ -136,6 +136,20 @@ class Main:
             if result == False:
                 # The item failed to add to the menu
                 xbmcgui.dialog().ok( __addon__.getAddonInfo( "name" ), __language__(32091) )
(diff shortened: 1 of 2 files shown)


ghost:
Should have commented I was going to wait to see if my PR to the library node editor was likely to get merged before this did (otherwise it seems wise to remove the integration code, rather than add to it ;)) but thanks for the merge :)

PR #70, opened 2015-09-23
[version] Bump version and changelog

BigNoid:
This bumps master to 0.6.1 and adds changelog entries for the missing version bumps.


marcelveldt:
Great, thanks!

What should we do with informing all those mod skins out there? When we removed the legacy support, we informed all known/official skinners but no way to track down all mod variants.
Maybe Ronie can post about it on his "Skin changes for ..." thread ?

Or just let those non official skinners find out the hard way that something has changed ?

ghost:
Indeed, my compliments @BigNoid for a both a nice and quick solution to the issue when it came to light.

As regards informing modders, I've reached out to the one modder who has comments on his thread about the disappearing submenus, and he'll fix it this weekend.

Issue #57, opened 2015-09-01
FR: customize property dialog


marcelveldt:
A feature request to be considered, submitted by tomer953 on the Kodi forums:
http://forum.kodi.tv/showthread.php?tid=178294&pid=2087427#pid2087427

Some requests about the Custom property method.
1. Custom Title for the choosing dialog
2. Default value for the property, so Every new item will have the default value
3. option to remove the "none" option
   
   for example, I have:

```
    <property property="Icon_Overlay">31140</property> <!-- Same as Text color -->
     <property property="Icon_Overlay">31142</property> <!-- Same as Skin color -->
     <property property="Icon_Overlay">31141</property> <!-- Original Icon Color (no overlay) -->
```

http://i.imgur.com/jTz7Yw5.png

 and I want to add custom title to the dialog, and set default value to 31142,so maybe adding new feature like: 

```
<property id="Icon_Overlay" title="Custom Title" default="31142"/>
```

or something similar.
 Also in that case, sometimes when I send click to "404" for one custom property (lets say choose color) the choose property dialog for other button (choose icon color) is opening right after the first one .
 so I'm getting dialog after dialog when I just requested one of them. but If Ill have custom title I'll wont care, because the user will know what he is setting.

ghost:
I'm not overly familiar with the code, but some of these sound both very sensible and (without looking into the code!) not too likely to cause breakages, if someone wanted to look at them before a stable push.
- Custom title - if only a single dialog is ever used, an override. Otherwise, done via a property set before sending a click?
- Default value - no comment, don't know the code well enough ;)
- Option to remove 'None' value - Same comments as for custom title.

PR #71, opened 2015-09-24
Custom property improvements & minor fixes

ghost:
Improvements to the ability to set custom properties
- New `<propertySettings />` element - to set the dialog title, whether ‘None’ will be shown and to enable image browsing
- buttonID attrib in propertySettings, to tell script to handle additional button to set the custom property (no more having to set a window property then send a click to 404)
- New label attrib for `<property />` element - to set label to display separately from value (will also translate $INFO’s) and condition attrib to decide if property should be shown
- Localise strings
- Clear window property ‘chooseProperty’ after sendClick to 404
- Update documentation

This has been tested by @Jayz2K via the Kodi forums, and should implement most of the requests in #57 (except default value for new items)

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -1,13 +1,14 @@
 script.skinshortcuts was written with the intention of making user customizable shortcuts on the home page easier for skinners.
 
 
-What's New for Skinners (version 0.5.4 - git version)
+What's New for Skinners (version 0.6.1 - git version)
 -----------------------
 
 - Ability to include multiple $SKINSHORTCUT[] elements and other text in an element
 - New <content>addon-program-plugin</content> for groupings and related widget-selection support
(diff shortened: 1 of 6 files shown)


ghost:
I've added a small fix if that's OK, rather than opening a new PR for it. As reported on the forum (or will be reported, when they sort out their debug log), it fixes a condition whereby the action can be blank if the user edits the shortcut. This fix ensures that if the custom action is empty, it gets set to "noop".

ghost:
Again, doesn't seem worth opening a PR for one line of code, so another commit which - finally and completely - fixes cases where one profiles properties can be applied to another profiles menus, by clearing the cache of custom properties when the profile we're building the menu for changes.

marcelveldt:
This is looking fine to me. It has also been reported as working and not breaking stuff :-)

ghost:
Cheers :)

PR #79, opened 2015-10-07
Fixes and requests

ghost:
This PR is a combination of minor fixes and (mostly) minor feature requests. @marcelveldt do you mind taking a look through when you have chance :)
- Fix auto-enabling debug logging - something appears to have changed in JSON, so the call being used to check whether system debug logging was enabled was failing.
- Just Select Widgets - widget version of the classic Just Select method
- Let users edit widgetName after setting property, along with override for skinners to stop it happening
- Officially let `<propertySettings />` use a Skin Shortcuts control as the buttonID - minor changes to behaviour and code cleanup so this should happen in a good way
- Let listitems without a label (but with an action) save - allows me to revert the Change Action button forwarding the click to Select Shortcut when there was no label
- Add optional icon element to `<property />`'s, along with a choice of ugly or pretty select dialog dependant on whether any have been specified
- Fix labelID's being generated twice by management dialog when overriding icons. Not sure it actually caused any issues, but helped when coding...
- Add ability to use 308 to restore a deleted shortcut, whilst still being able to reset all shortcuts if desired
- And finally, fix a rare condition (probably only experienced when messing with the code) whereby the management dialog could be permanently marked as 'loading' but failed to do so, and so couldn't be accessed

I know it's a lot of features, but each one is (purposefully) not a lot of code, so hopefully not too much chance of breakages!

Edit:- One more fix, reported on forum since the PR
- Incorrect code when parsing global overrides - the code was checking that there weren't any `<action />` elements, but then trying to apply the text from one. Updated docs too to highlight that (now) this element is not required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ script.skinshortcuts was written with the intention of making user customizable 
 - Ability to assign menu item ID to a property when using templates
 - New <content>addon-program-plugin</content> for groupings and related widget-selection support
 - More options for setting a custom property - [Read More](./resources/docs/advanced/Custom shortcut properties.md#set-options)
+- New 'Just Select Widget' method - [Read More](./resources/docs/advanced/Just Select.md#just-select-widgets)
 
 #### Version 0.6.0 - repo version
 
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ script.skinshortcuts was written with the intention of making user customizable 
 - Behaviour change - skin will now ask users to edit widgetName after setting widget. [Read how to prevent](./resources/docs/advanced/Managing widgets.md#prevent-user-editing-widgetname)
 - Ability to use a Management Dialog buttonID to set custom property. [Read More](./resources/docs/advanced/Custom shortcut properties.md#notes)
 - Label change - GUI 308's label (32028) has changed from "Reset shortcuts" to "Restore shortcuts" to highlight additional functionality
+- Remove requirement to include an `<action>` element and the `group=` attribute in a global action override.
 
 #### Version 0.6.0 - repo version
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ script.skinshortcuts was written with the intention of making user customizable 
 - New <content>addon-program-plugin</content> for groupings and related widget-selection support
 - More options for setting a custom property - [Read More](./resources/docs/advanced/Custom shortcut properties.md#set-options)
 - New 'Just Select Widget' method - [Read More](./resources/docs/advanced/Just Select.md#just-select-widgets)
+- Behaviour change - skin will now ask users to edit widgetName after setting widget. [Read how to prevent](./resources/docs/advanced/Managing widgets.md#prevent-user-editing-widgetname).
 
 #### Version 0.6.0 - repo version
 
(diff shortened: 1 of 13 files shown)


marcelveldt:
the code looks fine so I've merged it. I've also built another beta release and published with my skin so some users can give it a real testdrive. Version number for the beta (on emby repo) is 0.6.2

ghost:
Much appreciated as always :)

Want me to bump git to 0.6.2 so that version numbers don't get confused on any future repo push? (Assuming no further updates - unlikely, though I'll only be PRing fixes for the foreseeable - could then be used as repo version; otherwise ready for a 0.6.3 beta/release)

marcelveldt:
Yeah, please bump the version... If you don't have time I can do so

ohoh, I am now testing the new version very well...
This just happened after I reset the shortcuts:

20:57:40 T:1972   DEBUG: script.skinshortcuts: Failed to write menu
20:57:40 T:1972   ERROR: Traceback (most recent call last):
20:57:40 T:1972   ERROR:   File "C:\Users\mvand\AppData\Roaming\Kodi\addons\script.skinshortcuts\resources\lib\xmlfunctions.py", line 92, in buildMenu
20:57:40 T:1972   ERROR:     self.writexml( profilelist, mainmenuID, groups, numLevels, buildMode, progress, options )
20:57:40 T:1972   ERROR:   File "C:\Users\mvand\AppData\Roaming\Kodi\addons\script.skinshortcuts\resources\lib\xmlfunctions.py", line 386, in writexml
20:57:40 T:1972   ERROR:     menuitem = self.buildElement( item, "mainmenu", None, profile[1], DATA.slugify( submenu, convertInteger=True ), itemid = itemidmainmenu, options = options )
20:57:40 T:1972   ERROR:   File "C:\Users\mvand\AppData\Roaming\Kodi\addons\script.skinshortcuts\resources\lib\xmlfunctions.py", line 682, in buildElement
20:57:40 T:1972   ERROR:     if onclick.text.startswith( "pvr-channel://" ):
20:57:40 T:1972   ERROR: AttributeError: 'NoneType' object has no attribute 'startswith'

I am looking into the issue myself now too but maybe you have a clue from the top of your head

ghost:
Hmmm .. that's a strange error, and means I haven't got the updates to the global overrides correct. The code there searches for any overridden action - setting them to the onclicks - and, if none are found, sets the standard action to the onclicks. The fact that it is failing, though, suggests the shortcut has neither an override or a default action :confused: - most likely reason being the overrides aren't being applied properly. If you can't spot the issue, let me see you overrides and I'll sort it.

Edit:- Though, of course you may spot the issue in the time it takes me to type that :wink:

ghost:
More specifically, looking at the code, there IS an onclick element (the code wouldn't be running if there wasn't), but it has no text. Most likely, from the code I pushed, the text element of an override isn't being set correctly,

marcelveldt:
I'm not at my PC atm but will look once I'm back. Been a long time ago since I coded that override stuff

ghost:
Spotted the error - I'm checking against the wrong thing when I'm checking if there are no override actions specified. :blush: I'll do a new PR with the fix and the version bump shortly.

Sorry about that.

ghost:
Can you please test https://github.com/Ignoble61/script.skinshortcuts/tree/Fixes-and-requests and see if it fixes the issue when you have time.

(I'm not 100% convinced it will because, though it does correct an error it's one that - from my own testing - would have actually caused a different error, so may not be ultimately responsible...)

If it does, I'll do a new PR. If not, I'll keep looking!

Edit: Spotted a typo which explains (in conjunction with the wrong check for whether there any any overrides) which explains everything. Will open a new PR (though this won't be quick - the revert means that my branch thinks its changes are already part of BigNoid/master...), but would still appreciate if you could test the branch I linked when you have time to make sure I'm not missing anything.

PR #81, opened 2015-10-09
Fixes and requests

ghost:
Same as #79 but this time hopefully without the bug and with a bump to 0.6.2 :) @marcelveldt if I can impose on you further, when you have the time can you test if the base branch - https://github.com/Ignoble61/script.skinshortcuts/tree/Fixes-and-requests - has the bug you reported fixed? Cheers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
-#Skin Shortcuts - 0.6.1
+#Skin Shortcuts - 0.6.2
 
 script.skinshortcuts was written with the intention of making user customizable shortcuts on the home page easier for skinners.
 
 
 ## What's New for Skinners
 
-#### Version 0.6.1 - git version
+#### Version 0.6.2 - git version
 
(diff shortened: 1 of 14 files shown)


marcelveldt:
working perfectly fine now :-)

ghost:
So glad I tracked it down! Thanks for not only the original bug report, but for testing, then reporting with enough info to track the issue down, then for testing the fix, then for merging the final fix. :)

Now, if users could manage just one-out-of-five :wink: ...

marcelveldt:
haha!
Thanks as always for your hard work and patience.
I have just pushed the 0.6.3 beta version (yeah I incremented it just to be sure) to the Emby repo where my Titan beta skin is hosted for some real life tests.

ghost:
;)

I'll push another update (unless you object within the next couple of minutes) with a further version bump and create a tag at that point.

PR #83, opened 2015-10-10
Fix clone options, add clone all option

ghost:
As reported (and tested) on the forum, fixes the clonebackgrounds and clonewidgets build options, and adds a cloneproperties option to get them all.


marcelveldt:
Was this already tested by a skinner ?

marcelveldt:
I have the answer already ;-)

PR #84, opened 2015-10-10
translate label for widgetrename

marcelveldt:


ghost:
Means if it's already set to a $INFO/$VAR, it will get translated and stuck as the translated version - maybe a quick check that the keyboard.getText != widgetName, so if the user doesn't change it it will stay as the $INFO/$VAR...?

Or is there no reason widgetName would be preferred to be kept as the $INFO/$VAR? (I haven't used such for widgetName's myself, so I'm happy to be led by you on this one :smile:)

marcelveldt:
Ah, i get your point... so I'll send the translated label to the keyboard dialog than..

marcelveldt:
See, the additional commit; The translated label will only be saved if the user actually changes it.
Also did another commit to hide the "create shortcut here" default entry when browsing the default widgets and smartshortcuts provided by the skinhelper script because it caused some confusion among users.

ghost:
Sensible change :) and code looks good.

PR #86, opened 2015-10-12
Personal changes for Conq mod main menu

ghost:
Fully understand if its preferred not to merge these changes - I'm not sure how useful they'll be to people in general, they're just the changes I've needed to get the main menu of my Conq mod looking how I want it to look. As such, I'm quite happy for this PR to be rejected and I'll maintain this on my own fork. Obviously easier for me if they do get merged, though ;)

Two changes included:-
- Write the id of each item as a property
- Ability to specify minimum number of main menu items (if there aren't enough, the main menu items get re-written with the exact same properties, but an increased id)

Reason for the change:-

I wanted to extend the main menu of my Conq mod to the full height of the screen, but fade the items as they get further away from the focus point. As it's a wraplist this required two things - ensuring that there are more items in the menu than there are shown in the wraplist, and being able to individually identify the items.

With that, I'm able to apply a conditional fade animation as such `<animation effect="fade" start="100" end="70" time="0" condition="StringCompare(Container(9000).ListItem(-3).Property(id),ListItem.Property(id))">Conditional</animation>`, giving the end result of:-

screenshot000

Thanks for your consideration of the changes.

### DIFF
--- a/default.py
+++ b/default.py
@@ -65,7 +65,7 @@ class Main:
             xbmcgui.Dialog().ok(__addonname__, line1)
             
         if self.TYPE=="buildxml":
-            XML.buildMenu( self.MENUID, self.GROUP, self.LEVELS, self.MODE, self.OPTIONS )
+            XML.buildMenu( self.MENUID, self.GROUP, self.LEVELS, self.MODE, self.OPTIONS, self.MINITEMS )
             
         if self.TYPE=="launch":
             xbmcplugin.setResolvedUrl( handle=int( sys.argv[1]), succeeded=False, listitem=xbmcgui.ListItem() )
@@ -234,6 +234,7 @@ class Main:
         
         self.NOLABELS = params.get( "nolabels", "false" ).lower()
         self.OPTIONS = params.get( "options", "" ).split( "|" )
+        self.MINITEMS = int( params.get( "minitems", "0" ) )
         self.WARNING = params.get( "warning", None )
         self.DEFAULTGROUP = params.get( "defaultGroup", None )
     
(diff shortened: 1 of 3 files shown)


ghost:
Having thought through these additional functions some more:-

The additional property shouldn't cause any issues.

The repeated listitems (with the increased ID) has the potential to cause issues purely from a code-maintenance issue (the additional listitems are being rebuilt, so any code changes for the primary building would need to be reflected in this secondary building). It should be quite easy to use the already-built listitems and modify their existing ID, which would eliminate any ongoing maintenance.

As such, I'll make the changes to this PR when I am able, and merge it unless anyone has any objections...?

marcelveldt:
OK, that's great. I'll wait for your final changes then.

ghost:
Much appreciated :)

I've updated the PR to show what I intend to do, but this code is completely untested and I largely suspect won't work just because I haven't looked into what item-type the code is using (I'm not in a position where I can test it in an actual environment for the next couple of days.) I'll test it properly when I return, and then merge the changes. Though if you have time to test yourself (and my apologies when it just doesn't work) that would be great :)

Even if/though it doesn't work, it gives a good idea of what I intend, though ;)

It includes an additional fix, though tbh I'm not sure where this has come from - I regularly rebase my branches, so this appears to be from a fix I applied but never PR'd - which marks that the menu was changed if the widget was removed. This, at least, is a change that should be merged before any repo update;)

ghost:
Final version of code, tested working by me and with an additional safety check to ensure no possibility of entering a loop. As there seem to be no objections, I'm going to go ahead and merge this.

PR #88, opened 2015-10-17
Change Behavior of ::ACTION:: in action override

tomer953:
[tomer953] Remove requirement for ::ACTION:: in action override to be
the only text in element


marcelveldt:
Thanks, I will merge this.

BTW: next time please do not increment the version number in your commit. 

Thanks,

Marcel

PR #89, opened 2015-10-22
Insert new menu in the current location

tomer953:
management dialog - Insert new menu in the current location instead of
add them at the end


ghost:
Code looks good, and I know from the forums you've confirmed this working. Cheers.

PR #92, opened 2015-10-26
Add Hebrew Support for Widget Names

tomer953:
Because of Kodi known bug of reversing RTL Languages strings in the
Keyboard dialog - we check if string is in Hebrew and reverse it if
true.


marcelveldt:
I'm not entirely sure that this should be merged tbh... Yes, you're right this will be a first step in getting some RTL compatability but it's just the top of the Iceberg. Support for RTL should be handled by Kodi natively and not in every single script seperately.

On the other hand this code is not really in the way of things and quite safe.

@Ignoble61 what are your thoughts ?

ghost:
As I said on the forum, I'm :-1: on working around a Kodi bug like this script-side (especially when the issue finally does get fixed in core ;) but I won't block it :)

However, if it were to go in then it would need to be expanded to every time the script gets text input from the user - so (from a quick search through the code) also GUI 305 and within the SelectShortcut() function in library.py. And, as that's three different locations across two different py files, it should probably be wrapped up in a general function (possibly in datafunctions.py) and called from there, rather than having the same code included multiple times.

Also, again if it is to be merged, you'll need to fixup your commits :)

ghost:
@phil65 @BigNoid You both are probably in a better position than marcelveldt or I to comment on whether scripts should be working around a Kodi RTL bug...?

phil65:
Entirely up to the script author. I personally would say yes as long as the implementation is clean because I have no idea if anyone is workin on improving RTL support atm.

tomer953:
Thanks everyone for your comments.
As I said in the forum, I Accept your opnions about why scripts not should handle with the software bugs, but since this is bug from day one, and no fix in the near future, I Think that the more people are enjoy this script is the better.
As Marcel Said, this fix is only concern to ppl who use RTL languages, and it quite safe. Also another scripts, like Phil's extendedinfo are using the same, and I tested them quite well.
I think you can trust me, as one of the only skinner in my country, that I will be the first one to PR the oppiste fix if\when kodi will fix that bug.

Ignobel, I added a new commit about fixing the problem in the GUI 305, and I think thats all.
In the SelectShortcut() its not going to be, because Actions is written in English Only, and this fix Canno't deal with both English and RTL Language togethger, (I will flip the English as well).
So It should stay as is in the custom action.

Thank you all guys, hope to see it in.

ghost:
@tomer953 code wise it looks fine, but the commits still need to be squashed :) (please ask if you're unsure on doing this)

@marcelveldt considering phil65's :+1: for this and tomer953's assurances he'll manage removing it when the time comes (including, I presume, the needed python api bump :wink:), any further thoughts on whether this should be merged?

tomer953:
mm, I'm new to github, and I'm using the website gui function, I did not find any option to control the commits, should I remove my fork, re-add it, then add a single commit and PR it again?

Thanks again

ghost:
Yes, squashing needs to be done via the git command line tools. Certainly don't open up another PR with a single commit until it's been decided whether the PR will get in - we don't want to loose the conversation here!

PR #94, opened 2015-10-28
RTL Support for keyboard dialog

tomer953:
As a workaround to a Kodi bug, which reverse string in RTL Languages, we
check if string is in RTL Language and reverse it order to display it
right.


ghost:
Discussion of PR here:- #92 

General consensus is to merge it :)

PR #97, opened 2015-10-29
Fix application of DATA.xml visibility when action overriden

ghost:
As reported on the forum, visibility conditions from .DATA.xml files aren't being applied to menu items when the item is overriden (e.g. action override). This removes the check as to whether the action has been overriden.

Note:- I need to test this further for regressions before it is merged (there will likely be a reason it's done as it is, so I need to check what that is and ensure I'm not breaking anything.) Therefore, I'll merge it once the patch has been tested and I'm sure it won't cause further issues :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
-#Skin Shortcuts - 0.6.4
+#Skin Shortcuts - 0.6.5
 
 script.skinshortcuts was written with the intention of making user customizable shortcuts on the home page easier for skinners.
 
 
 ## What's New for Skinners
 
-#### Version 0.6.4 - git version
+#### Version 0.6.5 - git version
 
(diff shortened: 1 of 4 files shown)


ghost:
Have thought through original intention of the check. It was for when an action was directly replaced via an action override, in which case it couldn't be sure that the original visibility condition would still be applicable to the new action.

As it is, the general usage by skinners for overriding actions has been to apply different functions dependant on condition, on top of the original function. As such, it makes sense to always apply the original condition. Therefore, I don't think this change of behaviour will cause (m)any issues for skinners.

I've added a behaviour change note to the Whats new for skinners section. Also done a version bump, as requested on the forum.

If no objections, and no issues with it occur to me, I'll merge this in the morning.

PR #104, opened 2015-11-16
Cleanup

ghost:
I've performed basic sanity checks on these changes (including ensuring that I haven't obviously broken Nox 5 for a change :wink:), but two of the changes stand a good chance of introducing breakages so they could do with a second eye cast over them before they're merged - if anyone is up for that - and having some good public testing before any future repo push is thought about :smile:
- Single entry point for loading available shortcuts - previously, each function that loads a group of available shortcuts used the same code on entry which managed it being loaded into memory just once, and keeping Kodi waiting when it hasn't quite loaded yet, as well as the same error handling (for those that were likely to fail). This brings those elements into a single shared function, and means the individual functions are now responsible for just loading the library data they need.
- Return an empty tree if no overrides.xml can be loaded. Previously we returned 'None', which then had to be checked for (and caused the pre-repo issues with the last version). By returning an empty tree we (largely) remove the need for that (checks still needed for a `tree.find()` to ensure that it hasn't returned None, but that should be being done anyway...)
- Reverts a change from enabling/disabling Kodi's debug logging from the inbuilt ToggleDebug function back to doing it via JSON, as there were reports that in some situations the debug logging wasn't being disabled correctly with the inbuilt function

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 v0.6.7
 - Enable 'Get More...' installation of widget providers
+- Cleanup
 
 v0.6.6
 - Template improvements
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 v0.6.7
 - Enable 'Get More...' installation of widget providers
+- Return to using json to enable/disable Kodi debug logging
 - Cleanup
 
 v0.6.6
--- a/resources/lib/datafunctions.py
+++ b/resources/lib/datafunctions.py
@@ -203,13 +203,12 @@ class DataFunctions():
             # group overrides: add an additional onclick action for a particular menu
             # this will allow you to close a modal dialog before calling any other window
             # http://forum.kodi.tv/showthread.php?tid=224683
-            if skinoverrides != None:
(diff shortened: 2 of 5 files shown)


marcelveldt:
The code looks fine to me. Maybe best to just merge it in and place beta version on my repo again (and maybe some others) so let some fanatic users do the testdrive ?

BTW: Code is 100% good but some code can be more "pythonic", e.g.:
"if backgroundName is not None: " can be changed to "if backgroundName:"

ghost:
Heh, the pythonic thing is something phil65 has pointed out a few times with my code :wink: - it's because my (very limited) other coding experience is from much stricter languages, so I'm used to being much more explicit about things than the pythonic way, and I stick to it because it's much more readable for me. Maybe one of these days I'll go through the whole script and 'python' it up :grin:

OK, I'll go ahead and merge it and if you're happy to put it on the Emby repo I'll keep an eye out for any bugs that come up in the days ahead. Thanks for looking it over :)

marcelveldt:
haha, don't worry about it. As long as the code is working fine :-)

I'll put the testversion on my beta repo at Emby

PR #107, opened 2015-11-20
Fix kodiwalk function

ghost:
The playlist issue that was recently highlighted was ultimately caused by an error in the kodiwalk function - it would only walk the first directory (and then the first sub-directory and so-on) that it came to. This is a minor re-write of that function. It brings it down to a single recursive function rather than two, and it only now returns files (previously it also returned the root and the directories - information that we didn't need) in order to simplify. Oh, and it now walks all the directories it finds :wink: 

@marcelveldt I'm again going to go ahead and merge this as the chance of breakage is very low, but would you consider updating the script on the Emby beta repo so it can get some wider testing? Cheers.

### DIFF
--- a/resources/lib/library.py
+++ b/resources/lib/library.py
@@ -39,39 +39,25 @@ def log(txt):
         except:
             pass
 
-def kodilistdir(path, walkEverything = False, stringForce = False):
+def kodiwalk(path, stringForce = False):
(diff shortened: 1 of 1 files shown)


marcelveldt:
Yep, will do so today.

PR #110, opened 2015-11-23
Fix case where no visible items in management dialog

### LINKED ISSUE #109: index out of range on edit submenu
http://forum.kodi.tv/showthread.php?tid=234547&pid=2166157#pid2166157

To be fixed on some rainy day..

ghost:
This should close #109 

Found and fixed an issue where no items would be shown in the management dialog (including no `<none>` item) when there are hidden items in the list. Also fixes case where additional property not added when there are no shortcuts selected by the user at all for a menu.

### DIFF
--- a/resources/lib/gui.py
+++ b/resources/lib/gui.py
@@ -252,15 +252,9 @@ class GUI( xbmcgui.WindowXMLDialog ):
         
         # Initial properties
         count = 0
+        visible = False
         DATA._clear_labelID()
         listitems = []
-        
-        # If there are no shortcuts, add a blank one
-        if len( self.allListItems ) == 0:
-            listitem = xbmcgui.ListItem( __language__(32013), iconImage = "DefaultShortcut.png" )
-            listitem.setProperty( "Path", 'noop' )
-            listitem.setProperty( "icon", "DefaultShortcut.png" )
-            self.allListItems = [ listitem ]
         
         for listitem in self.allListItems:
             # Get icon overrides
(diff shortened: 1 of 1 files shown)


marcelveldt:
Code looking fine.. One comment however: I saw you changed the default widgets for skinhelper and also included static. That might not be a good idea because the static widgets won't be included in every skin ?

marcelveldt:
Thanks btw for fixing this!

ghost:
Damn. I was trying to look into why Skin Helper has difficulty with xbmcplugin.addDirectoryItem on Android x86 (assuming the error in JayZ's log - `Attempt to use invalid handle 31` - is the ultimate issue, though Kodi doesn't make it clear that's actually coming from Skin Helper and there's no obvious cause for it) so was playing with a few different ways to call the widgets to see if I could recreate a situation where xbmcplugin.endOfDirectory is wrongly called. I'm playing around with some other code ATM but I'll be sure to revert that before this gets merged.

(And you're welcome, assuming it actually is fixed!)

ghost:
Fixed up code - I'd forgotten to convert an int to a str when adding the skinshortcuts-orderindex property to the blank listitem. (And the skin helper changes are removed.)

ghost:
Just had confirmation that this pr still doesn't fix the issue - http://xbmclogs.com/p7yzhcjue

Any ideas what else might be causing this?

(Edit:- Actually, the line numbers don't match up between the log and the latest PR, so it may still fix the issue. But if anyone _does_ have any other ideas as to the cause... ;0)

ghost:
As there doesn't seem to be any further feedback coming on whether this fixes the problem for the users who have reported it - and because I have thoroughly tested in the one situation where I could reproduce the issue, I'm going to merge at this time. Even if it doesn't fix the specific issue the user was experiencing (and it should), it does fix a bug and from testing doesn't cause any further issues.

Issue #22, opened 2015-07-03
Skin Shortcuts 1.0.0


ghost:
I had a list of tasks that needed to be done within Skin Shortcuts before I was happy to declare it a 1.0.0 release and, with Isengard around the corner, I thought it would be interesting to post it here - with a few notes - to see if the script could be got up to a 1.0.0 release to coincide with Kodi J*, whenever that lands. And - more importantly - to see what others think would be necessary for a 1.0.0 release, and where they disagree with my views :)
1. Most important, this script should work for all users without crashes. That doesn't mean there won't be errors (preferable, but lets be realistic!), but the script itself should work in normal circumstances (exceptional circumstances being a broken Kodi installation, user/skinner error, etc)
   - When I last engaged fully with the community there was still the occasional issue saving the menu. I would expect PR #21 to go someway towards tracking these down, if they still happen.
2. Remove legacy code / Other code cleanup. The script in its current form is nothing like the script I originally envisioned (to be brief, it wasn't originally intended to provide the whole main menu, just - well - a list of shortcuts). The fact that this didn't appear to be what the community wanted means the code developed as necessary - and as my skills (even more limited then than they are today) allowed - to support this and features I deemed necessary to do this + requested features were added ad hoc, rather than in a planned way. It would be very nice to remove a lot of the legacy code, which would clean up the code and improve performance - but would likely break some skins if the cleanup isn't done more carefully than I could probably manage!
   - Removing .shortcuts compatibility would break Nox5 + variants, and possibly others, but would be great.
   - Menu building code has become ridiculously overloaded and needs a revamp. This would almost certainly speed up menu building and simplify code, but has the potential to break skins if not done very carefully!
   - Ideally, store main menu/submenu/additional menus in a single file. This would go someway to getting rid of the horror that is the labelID system, but stands a very good chance of breaking - and breaking badly - if not done perfectly, and is a lot of work as it affects most every area of the script. And, of course, for a 1.0.0 release, it'd be great to drop support for the then-deprecated .DATA.xml files, too...
   - Best way may be to break backwards compatibility (supporting J\* would mean 4 versions of Kodi supported in one script version - which already causes some issues regarding which shortcuts are available on which version)...
[...]

PR #112, opened 2015-11-27
Version bump to mark stable

ghost:
As discussed in #22, bumps version to 1.0.0 and marks stable. To be merged ahead of repo push.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
-#Skin Shortcuts - 0.6.7
+#Skin Shortcuts - 1.0.0
 
 script.skinshortcuts was written with the intention of making user customizable shortcuts on the home page easier for skinners.
 
 
 ## What's New for Skinners
 
-#### Version 0.6.7 - git version
+#### Version 1.0.0 - repo version
 
 - Ability for users to install additional widget providers - [Read More on customising when this is available to users](./resources/docs/advanced/Managing widgets.md#default-widgets-from-skin-helper-service)
 - Jarvis changes to music shortcuts - Please be aware that the shortcuts used to access various areas of the music library have changed. Skin Shortcuts will do its best to update any actions to the Jarvis version when the menu is built, but if this doesn't work (or if you are using the 'Just Select' method) the user may need to reset the action for these shortcuts.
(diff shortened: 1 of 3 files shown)


marcelveldt:
Current repo version has been tested for over a week now on several 3th party repos, so far no issues reported. Go ahead with the repo push ?

ghost:
:-1: We don't want to break Nox 5 for once, so I just gave it a quick test and it looks suspiciously like the 'Just Select' method (used by Nox 5 for the main menu items) isn't correctly returning actions. Not sure why as I didn't think that code has been touched, but I'll debug this evening. Once thats fixed, though, and confirmed the rest of Nox 5 is working correctly, should be good for repo push.

ghost:
Ignore me - I'd forgotten I'd hacked Nox 5 to use the Just Select Widgets method so I could quickly test it :blush: 

Yes, I think we're good for a repo push.

ghost:
@marcelveldt you're obviously busy with real life at the moment. I've studiously avoided the Kodi mailing list since returning to Kodi scripting so far, but if you're too busy to deal with requesting the update on the repo, let me know and I'll face the fear and deal with it myself ;)

ghost:
I don't have time to do this until Tuesday eve at earliest, but the suggestion coming from the forum is that the script focuses GUI 211 when the shortcuts are loaded - potentially with an override allowing the skin to select 303 or 304 if that is appropriate to the skin - before a repo push. See conversation from http://forum.kodi.tv/showthread.php?tid=178294&pid=2186091#pid2186091 onwards.

marcelveldt:
I can/will do the repo push tonight, no problem. My internet connection got broken so I was offline for a couple of days :-)

About the focus of the control. I included some method for that in skinhelper.

RunScript(script.skin.helper.service,action=setfocus,control=201)

marcelveldt:
Decided to not include the focus of control in the script at this time because that could possibly break other use cases in other skins. Best to leave it as is for now and let the skinner decide what control to give focus: They can use the delayed focus method in skinhelper for that or just focus an already existing control in their xml.

BTW: Another valid method is to set the window onload to focus a hidden button.
Set the onfocus of that hidden button to perform any action you like. This will prevent errors in the log and is more fail safe.

Examples in confluence skin startup.xml or Titan's home.xml

PR #113, opened 2015-11-28
Add RestartApp Kodi command

ghost:
Adds 'RestartApp' to the available shortcuts for Kodi Commands. Whilst the wiki says this is available for Windows and Linux, it needs confirmation that this is also available on rPi (which is based on Linux, but often has different behaviours when is comes to this sort of action) - http://forum.kodi.tv/showthread.php?tid=249817. Once confirmed, this can be safely merged as there is essentially no chance of breakage.


ghost:
Confirmed RestartApp is not available on rPi. PR updated to reflect this.

PR #114, opened 2015-11-29
Fix music add-on shortcuts on Isengard

ghost:
As reported here - http://forum.kodi.tv/showthread.php?tid=210069&pid=2173464#pid2173464 - and confirmed by myself, shortcuts to music addons don't work on Isengard (they display a black screen.) Not 100% sure why this should be, as they link to window 10501 (MusicFiles) which I understood to be available on Isengard :question:, but changing it to 10502 (MusicLibrary) fixes the problem, and is forward-compatible with Jarvis (and shouldn't get hit by any auto-visibility conditions requiring there to be a music library).

### DIFF
--- a/resources/lib/library.py
+++ b/resources/lib/library.py
@@ -1472,7 +1472,7 @@ class LibraryFunctions():
                 listitem = xbmcgui.ListItem(label=label[ len( label ) - 1 ].replace( "  >", "" ), label2=localItemType, iconImage="DefaultShortcut.png", thumbnailImage=thumbnail[ len( thumbnail ) - 1 ])
                 
                 # Build the action
-                if itemType == "32010" or itemType == "32014" or itemType == "32069":
+                if itemType in [ "32010", "32014", "32069" ]:
                     action = 'ActivateWindow(10025,"' + location + '",return)'
                     listitem.setProperty( "windowID", "10025" )
                     listitem.setProperty( "widgetType", "video" )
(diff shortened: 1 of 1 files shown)


resources/lib/library.py
```diff
@@ -1494,8 +1494,8 @@ def explorer( self, history, location, label, thumbnail, itemType, isWidget = Fa
                     listitem.setProperty( "widgetPath", location )
 
                 elif itemType == "32011" or itemType == "32019" or itemType == "32073":
```
phil65: `elif itemType in ["32011", "32019", "32073"]:` might look a bit cleaner. :)

resources/lib/library.py
```diff
@@ -1494,8 +1494,8 @@ def explorer( self, history, location, label, thumbnail, itemType, isWidget = Fa
...
-                    action = 'ActivateWindow(10501,"' + location + '",return)'
-                    listitem.setProperty( "windowID", "10501" )
+                    action = 'ActivateWindow(10502,"' + location + '",return)'
+                    listitem.setProperty( "windowID", "10502" )
```
phil65, line 1498: in general there should be no need to mess with the window ids at all. The preferred way for calling / referencing windows is by using their alias ("musiclibrary" in this case)

ghost, line 1498: Does it make a practical difference? The decision to use window id's for add-ons is because the script doesn't apply any visibility conditions to ActivateWindow([windowid],...) whereas it does to the alias - and with this fix music add-ons wouldn't show unless there is music scanned into the library/the visibility conditions are made more complex...

phil65, line 1498: no, it doesnt make a practical difference, but it's more readable and less likely to get changed in the future.
Just a recommendation, if you want to stick with ids then feel free to do so.

ghost, line 1498: As this is a quick fix, I'll leave it as-is for now and rework to window aliases when I have a bit more time. Thanks as always for the comments :)

ghost:
Cleaned up code as suggested by @phil65, with suggestion to replace window id's with window aliases to go in after repo push (along with necessary visibility condition changes).

ghost:
Tested as thoroughly as I can. @marcelveldt is there still time to get this in (with a small chance of a breakage due to its nature) before final testing on the Emby repo?

marcelveldt:
Yep, just merge the code. I will then put a beta version on my beta repo to let some actual users take it for a testdrive.

ghost:
Will do. Cheers :)

PR #115, opened 2015-11-30
Ability to set widgetName separately from label

ghost:
As requested on the forums.


ghost:
Confirmed working via forums.

PR #117, opened 2015-12-02
Fixes - clones properties, templates

ghost:
Fixes two issues reported on the forums
- Cloned property values weren't being converted from DATA.local() format to actual localised format (confirmed fixed)
- Code that searches for 'other' templates could return early

Second one may fix issue reported on forum - waiting for confirmation. If not, will continue looking into that problem, and try and get all fixes in on one PR ;)


ghost:
Template update confirmed as fixing issue.

PR #118, opened 2015-12-04
fix: widgetname not transferred from skinhelper

marcelveldt:


ghost:
I can't really comment on the code as I've never looked into exactly what Skin Helper is returning at that point, but assuming you've tested this fixes the issue (and I'm sure you have), go ahead and merge :)

PR #119, opened 2015-12-06
Template / single build mode fix

ghost:
As reported on the forum, templates visibility condition doesn't match to submenu items when using single build mode. Requesting original reporter to confirm fix.


ghost:
Fix confirmed.

PR #124, opened 2015-12-29
Adjust Isengard + earlier music addon shortcut dependant on music lib…

ghost:
…rary on Isengard and earlier

As originally reported:- http://forum.kodi.tv/showthread.php?tid=252302, with condition for action provided by Ronie here:- http://forum.kodi.tv/showthread.php?tid=178294&pid=2198065#pid2198065

Needs confirming that non-working music add-ons are being accessed via a Music Add-ons shortcut, and that the fix works. Worth considering a repo push to get this fix out before Isengard becomes old news...?

### DIFF
--- a/resources/lib/datafunctions.py
+++ b/resources/lib/datafunctions.py
@@ -1043,7 +1043,31 @@ class DataFunctions():
     def upgradeAction( self, action ):
         # This function looks for actions used in a previous version of Kodi, and upgrades them to the current action
 
-        # Jarvis music changes
+        # Isengard + earlier music addons
+        if int( __xbmcversion__ ) <= 15:
+            # Shortcut to addon section
+            if action.lower().startswith( "activatewindow(musiclibrary,addons" ) and xbmc.getCondVisibility( "!Library.HasContent(Music)" ):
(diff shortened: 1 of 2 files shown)


ghost:
Thinking through this, this will only affect the shortcut at menu build time - if the boolean condition changes after that, the menu item will not update. Therefore this either needs handling in a different way (at present there's no mechanism to add a second action to a shortcut dependant on what the action is, which is what would be needed...) or we need to rebuild the menu on the boolean condition change (can be handled with the .hash file).

ghost:
Updated to also adjust action for shortcuts to specific music addons based on whether there is music content, and to rebuild the menu if music content has changed. (All Isengard only)

marcelveldt:
Is this ready to test ? Sorry for the late feedback. I was ill for the last couple of days and besides that busy with family.

BTW: Happy new year to you and your family

ghost:
I'd like to confirm whether the changes to shortcuts directly to music addon nodes are needed - but it's the wrong time of year for this sort of thing so haven't got round to testing it yet :)

Hope you're feeling better, and that 2016 brings great things to you and yours.

PR #126, opened 2016-01-13
Mark Stable

ghost:
Updates documentation to mark stable; version bump.

Also removes xbmc.sleep() which I understand is a requirement for the repo these days :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,16 @@
-#Skin Shortcuts - 1.0.0
+#Skin Shortcuts - 1.0.2
 
 script.skinshortcuts was written with the intention of making user customizable shortcuts on the home page easier for skinners.
 
 
 ## What's New for Skinners
 
-#### Version 1.0.1 - git version
+#### Version 1.0.2 - repo version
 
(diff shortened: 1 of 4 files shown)


marcelveldt:
is this ready for the official repo ? Or testdrive on beta repo for a few days ?

ghost:
It should be safe to go straight to the official repo if you're comfortable with that. The Isengard music fixes have been tested quite well by various people who have experienced the issue. Of the other changes, loading default properties correctly is a simple fix for an annoying bug; replacing xbmc.sleep shouldn't cause any issues and the other changes are for an in-development update to skin, so even if they don't work properly they shouldn't actually cause any issues outside of the skin that will use them :)

I'm also planning to PR the UX branch next week if everything goes to plan, which stands a good chance of breaking things. So getting a stable release out first would be good.

(And I've tested that nothing has broken Nox 5!)

marcelveldt:
OK Pull request emailed to the Kodi devs.
I'll let you know once it's accepted.

PR #128, opened 2016-01-19
Fix artist detectPluginContent error

ghost:
When `item["artist"]` isn't an array, detectPluginContent causes an error. This adds a check for an array and, if its not, checks against the value of `item["artist"]` directly.

@marcelveldt this is your code, so are you happy with this fix?


marcelveldt:
looks fine to me.

PR #129, opened 2016-01-20
End user-focused changes

ghost:
These commits are various changes to the script primarily aimed at the end users, unlike most commits which are normally requested by skinners.
- Context menu extension point (enabled from settings) to add any folder directly to the home menu, including auto-filling the submenu
- Language changes, moving away from referring to 'shortcuts' towards 'menu items', also should clear up some functions (choose shortcut/change action)
- Add option not to hide PVR shortcuts (requested feature)
- Use mtime, not hash of file, to determine if .DATA.xml/property files/overrides/etc have changed - should reduce disk access significantly
- Oft requested feature, the ability for users to specify that their menus aren't shared amongst skins utilising the script, plus ability to import menu from another skin
- Fix for disappearing icons between skins - icons and thumbs are now saved in the skins .properties file

Also bumps version and docs to mark unstable.

Context menu required xbmc.python 2.20.0 (Isengard), but I believe the script is on the Isengard repo...?

The non-sharing skins feature requires changes to skin.helper.service's backup functions, which I will PR separately.

The language changes will probably only be effective if the script is added onto Transifex.

@marcelveldt do you have time to give these changes a look through?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,20 @@
-#Skin Shortcuts - 1.0.2
+#Skin Shortcuts - 1.0.3
 
 script.skinshortcuts was written with the intention of making user customizable shortcuts on the home page easier for skinners.
 
 
 ## What's New for Skinners
 
+#### Version 1.0.3 - git version
+
+- Icons and thumbs are now saved on a per-skin basis
+
+Also, whilst it does not affect how skinners interact with the script, it is important to note this version introduces an option for users to specify that their menus should not be shared between skins.
+
 #### Version 1.0.2 - repo version
 
 - Ability to match multiple values when setting a property in a template - [Read More](./resources/docs/advanced/Templates.md#set-a-property-based-on-the-value-of-a-main-menu-item)
 - Ability to specify an 'onback' value for specific controls in management dialog - [Read More](./resources/docs/started/Management Dialog.md#custom-focus-on-back)
-
-#### Version 1.0.0 - previous repo version
-
-- Ability for users to install additional widget providers - [Read More on customising when this is available to users](./resources/docs/advanced/Managing widgets.md#default-widgets-from-skin-helper-service)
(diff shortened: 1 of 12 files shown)


phil65:
Would be nice if you could move the context menu extension to a separate add-on. That way users can choose if they want it (otherwise the context menu becomes very cluttered quickly)

ghost:
It's currently a setting and disabled by default, in an attempt to avoid having to actually request a new add-on to the repo ;) but I'll move it into a separate add-on tomorrow :)

phil65:
Ah ok, that might also work. I think it´s nicest though if all additional entries are listed in context menu section from add-on manager. If you agree: cool. If not: also fine.

ghost:
Moved context menu to separate addon - https://github.com/Ignoble61/context.skinshortcuts.addtomenu - as @phil65 suggests.

guidosarducci:
@Ignoble61  These are very welcome changes -- thank you!  I had previously been trying out the non-shared menu changes from your UX branch, but suggest making non-shared menus the default addon setting since it is much less frustrating for users.

For example, a common case is when a long-time skin user tries to test-drive a new skin, does some simple configuration, but then returns to the old skin to find the original menus messed up.  I also want to mention that in the 4-5 skinshortcut-enabled skins where I've tweaked menus to my liking, the menus ended up different between the skins, again making non-shared menus preferable.

ghost:
It will not be enabled by default in this version as that would cause everyone's menus to reset to default. It is something that could be discussed for a future version as the setting would then only take effect for new installations.

If people are experiencing menu's being "messed up" it is more preferable to post a debug log + screenshots of the issue so it can be fixed than for Skin Shortcuts to just disable the shared menu's by default. The only issue with shared menu's I'm currently aware of (which is to say has been reported on the forums) is with icons disappearing between skins, and this branch contains a fix for that issue.

PR #130, opened 2016-01-22
Fix m3u playlists + support video m3u's

ghost:
As reported by Braz on the Kodi forum, listing of m3u playlists was broken when the KodiWalk function was re-written. Also extends support to video m3u playlists - previously all m3u's were presumed to be audio playlists.

### DIFF
--- a/resources/lib/library.py
+++ b/resources/lib/library.py
@@ -47,7 +47,7 @@ def kodiwalk(path, stringForce = False):
     if json_response.has_key('result') and json_response['result'].has_key('files') and json_response['result']['files'] is not None:
         for item in json_response['result']['files']:
             if item.has_key('file') and item.has_key('filetype') and item.has_key('label'):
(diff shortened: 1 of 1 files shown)


resources/lib/library.py
```diff
@@ -47,7 +47,7 @@ def kodiwalk(path, stringForce = False):
     if json_response.has_key('result') and json_response['result'].has_key('files') and json_response['result']['files'] is not None:
         for item in json_response['result']['files']:
             if item.has_key('file') and item.has_key('filetype') and item.has_key('label'):
-                if item['filetype'] == 'directory' and ((not item['file'].endswith('.xsp')) and (not item['file'].endswith('.xml/')) and (not item['file'].endswith('.xml'))):
+                if item['filetype'] == 'directory' and ((not item['file'].endswith('.xsp')) and (not item['file'].endswith('.m3u')) and (not item['file'].endswith('.xml/')) and (not item['file'].endswith('.xml'))):
```
phil65, line 50: endswith can also take a tuple as first argument ;-)

ghost, line 50: Sure, but where's the fun in that ;) Next time I'm changing code near that, I'll update it as you suggest. Many thanks as always Phil.

PR #134, opened 2016-01-30
Updates

ghost:
After the end-user focused updates, a few updates more focused at skinners :)
- Support for building variables in templates
- Conditional property fallbacks - allows custom property fallbacks to be assigned based on other properties of the menu item
- Never pass additional properties through DATA.local function - as they aren't shared between skins, they don't need to be stored in a special format, and this also allows numeric values in additional properties
- Use window aliases rather than IDs - as suggested by Phil
- Build and management dialog speed improvements - build improvements mostly come from manually cloning elementtree objects, rather than copy.deepcopy'ing them. Management dialog improvements come from moving the loading of vfs images to a background thread so they don't block the loading.

@marcelveldt I know you're busy with real life, but do you have time to look through the changes. No worries if not, and hope things are well with you :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,14 @@ script.skinshortcuts was written with the intention of making user customizable 
 #### Version 1.0.3 - git version
 
 - Icons and thumbs are now saved on a per-skin basis
+- Ability to specify an icon to be displayed with a background - [Read More](./resources/docs/advanced/Managing backgrounds.md#defining-backgrounds)
+- Support building variables in templates - [Read More](./resources/docs/advanced/Templates.md#other-template)
+- Ability to specify conditions for fallback custom properties - [Read More](./resources/docs/advanced/Custom shortcut properties.md#for-all-shortcuts-without-a-user-set-property)
+
+Whilst they do not affect how skinners interact with the script, it is worth noting this version introduces changes aimed at the end users. Most notably:-
+
+- Users can now specify that their menus should not be shared between skins
+- Support for context menu add-on ([Git](https://github.com/Ignoble61/context.skinshortcuts.addtomenu)) to add items directly to menu
 
 Also, whilst it does not affect how skinners interact with the script, it is important to note this version introduces an option for users to specify that their menus should not be shared between skins.
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,7 +2,12 @@ v1.0.3
 - Option not to share menus between skins; never hide PVR shortcuts
 - Icons and thumbs are saved on a per-skin basis
 - mtime of files is used to determine if they have changed
+- New icon element for background overrides
+- Support building variables in templates
+- Add conditional custom property fallback override
 - Various language changes
+- Speed improvements
+- Minor fixes
 
 v1.0.2
 - Fix for music addon shortcuts on Isengard
--- a/resources/docs/advanced/Custom shortcut properties.md
+++ b/resources/docs/advanced/Custom shortcut properties.md
@@ -66,13 +66,21 @@ These will be applied when the user first switches to your skin, or when they re
 
 You can set what value should be used for all shortcuts in a given group, if the user hasn't selected an alternative value:-
 
-`<propertyfallback group="[GroupName]"" property="[Property]">[Property Value]</propertyfallback>`
+`<propertyfallback group="[GroupName]"" property="[Property]" attribute="[Attribute]" value="[Value]">[Property Value]</propertyfallback>`
 
 | Property | Optional | Description |
 | :------: | :------: | ----------- |
 | `[GroupName]` | Yes | The group that the labelID must be in, for example "movies". If omitted, the property will be apply to items in the main menu. |
 | `[Property]` | | The property of the shortcut that will be set |
+| `[Attribute]` | Yes | If property is conditional, the attribute that will be matched against |
+| `[Value]` | Yes | If property is conditional, the value that the attribute specified must have |
 | `[Property Value]` | | The default value of the property |
+
+You can include multiple `<propertyfallback />` elements for a specific property. The property will be set to the first one matched.
+
+When using conditions, both the [Attribute] and [Value] must be specified. So, if you wanted to match the fallback to a shortcut with the attribute 'widgetType' set to the value 'movies':-
+
+`<propertyfallback property="customProperty" attribute="widgetType" value="movies">Fallback Value</propertyfallback>`
 
 Note:- This will be applied to all shortcuts when the menu is built, and will show in the management dialog, but the fallback property will not be saved as part of the skins .properties file.
 
--- a/resources/docs/advanced/Templates.md
+++ b/resources/docs/advanced/Templates.md
@@ -69,6 +69,11 @@ The `<controls />` element must be included. Put your gui xml within.
 			</kodi>
 		</your>
 	</controls>
+	<variables>
+		<variable name="customvariable">
+			<value condition="SKINSHORTCUTS[propertyName]">$SKINSHORTCUTS[propertyName]</value>
+		</variable>
+	</variables>
 </other>
 ```
 
@@ -81,7 +86,9 @@ The `<controls />` element must be included. Put your gui xml within.
 | `<skinshortcuts>visibility</skinshortcuts>` | Yes | This will be replaced with a visibility condition based on the focus of the main menu |
 | `$SKINSHORTCUTS[propertyName]` | Yes | This will be replaced by the value of a matched <property /> element |
 
-The `<controls />` element must be included. Put your gui xml within.
+The `<controls />` element is where you put your Kodi GUI xml that should be built by the template.
+
+The `<variables />` element is where you define any variables that should be built by the template. At least one of them must be included. You must include the `<variable>` element, with as many `<value>` sub-elements as needed.
 
 One Other template will be built for each main menu item per [include]. Use different [include]'s to build multiple Other templates.
 
--- a/resources/language/German/strings.po
+++ b/resources/language/German/strings.po
@@ -7,14 +7,16 @@ msgstr ""
 "Project-Id-Version: XBMC Addons\n"
 "Report-Msgid-Bugs-To: alanwww1@xbmc.org\n"
 "POT-Creation-Date: YEAR-MO-DA HO:MI+ZONE\n"
-"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"PO-Revision-Date: 2016-01-30 10:04+0100\n"
 "Last-Translator: Kodi Translation Team\n"
-"Language-Team: German (http://www.transifex.com/projects/p/xbmc-addons/language/de/)\n"
+"Language-Team: German (http://www.transifex.com/projects/p/xbmc-addons/"
+"language/de/)\n"
+"Language: de\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Language: de\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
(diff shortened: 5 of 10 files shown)


marcelveldt:
Wow, looks like you had some time left ;-)
Yeah, all is fine here. Busy with family but also working on some rework of my skin that I needed to do some time.

I will take a look at your changes, merge them and test them.
Also I think it's a good idea to put a beta build on the beta repo again

ghost:
All the result of being involved in some skinning for the first time in ages - it's always amazing how much that highlights this scripts weaknesses ;)

Beta repo would be a very good idea - between these and the last UX changes, I think the chances of some breakages are high. Would it be at all possible to get the context menu add-on on the beta repo too, as the two will work together in the future?

marcelveldt:
Quickly reviewed the code, looks fine to me but I will have to do some stress testing to confirm. Nice idea to preload the backgrounds in a seperate thread to speedup the loading time. I see one potential issue though (but I could have read the code wrong) because there a unlimited while loop waiting for the backgrounds to populate but what if there's an error ? That would lockup the script, right ?

As I am currently doing a lot of testing with my Titan skin (which utilizes almost everything the script has to offer) after my major rework I can also perform the testing of the script changes and putting it on the beta repo will be even more valuable with real users testing the impossible scenarios :-)

About the contextmenu addon. Sure I will put that on the repo too, no problem.
Also I was thinking, why not include it in the script itself and make it an optional setting ? That would prevent users from having to install another addon and you still ahve control over what's displayed in the contextmenu because of the setting.

ghost:
It's a good point on the loop. I'll put a (shall we say) max 2 second wait on it which realistically should be more than enough.

Having it built in and done via an option was my first choice (and first implementation). Two problems with it - firstly, @phil65 convinced me a separate add-on was the right way (see #129). Second, when the script reloads the skin after building the menu, Kodi doesn't seem to correctly parse whether the setting is enabled or not, and so the context menu disappears - by doing it in a separate addon, that's not an issue.

marcelveldt:
Ah, yes that makes sense. After testing that should also be submitted to the Kodi repo, right ?
I will merge in the changes and start testing.

I think I already have a use case for the variables in the templates because I'm figuring out a way to force refresh some widgets that users select.

ghost:
The idea is it will be on the repo when its fully tested, yes.

The variables are, I think, going to be very useful. Haven't played with them myself, but seeing some of the stuff being done by Mr. V and his widgets is quite remarkable :)

Cheers for the review, the merge and the testing!

marcelveldt:
Bad news... the script keeps crashing...

IOError: [Errno 2] No such file or directory: 'C:\Program Files (x86)\Kodi_Isengard\portable_data\userdata\overrides.xml'

Keeps building the menu over and over. Seems to me that it is trying to locate the override file at the wrong location ?

marcelveldt:
FYI: I use a portable install of Kodi so I can have multiple versions installed at the same time

ghost:
What line?

marcelveldt:
00:19:55 T:8992   ERROR: Traceback (most recent call last):
00:19:55 T:8992   ERROR:   File "C:\Program Files (x86)\Kodi_Isengard\portable_data\addons\script.skinshortcuts\resources\lib\datafunctions.py", line 462, in _get_overrides_user
00:19:55 T:8992   ERROR:     tree = xmltree.parse( xbmc.translatePath( overridePath ) )
00:19:55 T:8992   ERROR:   File "C:\Program Files (x86)\Kodi_Isengard\system\python\Lib\xml\etree\ElementTree.py", line 1182, in parse
00:19:55 T:8992   ERROR:     tree.parse(source, parser)
00:19:55 T:8992   ERROR:   File "C:\Program Files (x86)\Kodi_Isengard\system\python\Lib\xml\etree\ElementTree.py", line 647, in parse
00:19:55 T:8992   ERROR:     source = open(source, "rb")
00:19:55 T:8992   ERROR: IOError: [Errno 2] No such file or directory: 'C:\Program Files (x86)\Kodi_Isengard\portable_data\userdata\overrides.xml'

marcelveldt:
I think you should a xbmcvfs.exists in there and correct the path

ghost:
The lack of the file is handled there in a try/catch block - that shouldn't be causing a build loop. Assuming skin shortcuts debug logging is enabled, why does it say it is rebuilding the menu each time? (The bug will likely be due to the change of using the file time)

(The file path is correct, btw - that's where the users own overrides.xml file would be if they supplied one. The lack of xbmcvfs.exists is intentional, though up for review - it's actually quicker to let it fail than to check if the file exists first)

marcelveldt:
I'll enable debug logging. I'm testing on a windows system btw

marcelveldt:
Nothing shows up additionally in the log when I enable debug logging so I guess the issue lies in the change detection. I'll take a look at the code

marcelveldt:
I think you should replace os.path.getmtime( filename ) with the xbmcvfs equivalent

ghost:
The log will say why it's rebuilding the menu - the line after it lists the profiles. For example, when there's no .hash file:

`23:34:39 T:123145316945920   DEBUG: script.skinshortcuts: script version 1.0.3 started`
`23:34:39 T:123145316945920   DEBUG: script.skinshortcuts: Profile found: Master user (special://masterprofile/)`
`23:34:39 T:123145316945920   DEBUG: script.skinshortcuts: Profile found: Test (profiles/)`
`23:34:39 T:123145316945920   DEBUG: script.skinshortcuts: No hash list`

Certainly can change it to the equivalent function. Give me a minute or two ;)

marcelveldt:
Hmmm, I replaced **os.path.getmtime(filename)** with **xbmcvfs.Stat(filename).st_mtime()** and still no go.

marcelveldt:
00:37:05 T:760   DEBUG: script.skinshortcuts: C:\Program Files (x86)\Kodi_Isengard\portable_data\addons\skin.titan\shortcuts\overrides.xml has been modified

ghost:
That's odd. It's obviously generating the mtime wrongly for the skins overrides either when writing the .hash file or when reading it. And I can't just see why that would be looking at the code :confused:

marcelveldt:
Ah, got it... added a little print statement to xmlfunctions:

modTime = xbmcvfs.Stat(hash[ 0 ]).st_mtime()
                        print "modtime --> " + repr(modTime)
                        print "hash --> " + repr(hash[ 1 ])
                        if modTime != hash[ 1 ]:

result:
00:40:52 T:11112  NOTICE: modtime --> 1454026691L
00:40:52 T:11112  NOTICE: hash --> 1454026691.613209

ghost:
What do you get if you revert that back to os.path.getmtime()? (It makes sense that os.path and xbmcvfs won't necessarily return the data formatted the same)

(The next line of the scripts debugging will show the old vs the new mtimes, btw :wink:)

marcelveldt:
got this when I reverted back to os.path.getgmtime:

0:55:16 T:12396  NOTICE: modtime --> 1454026691.613209
00:55:16 T:12396  NOTICE: hash --> 1454026691.613209
00:55:16 T:12396  NOTICE: modtime --> 1454196859.948804
00:55:16 T:12396  NOTICE: hash --> 1454196859.948804
00:55:16 T:12396  NOTICE: modtime --> 1448998656.62317
00:55:16 T:12396  NOTICE: hash --> 1448998656.62317
00:55:16 T:12396  NOTICE: modtime --> 1448998656.62317
00:55:16 T:12396  NOTICE: hash --> 1448998656.62317
00:55:16 T:12396  NOTICE: modtime --> 1446508045.0788693
00:55:16 T:12396  NOTICE: hash --> 1446508045.0788693
00:55:16 T:12396  NOTICE: modtime --> 1446508045.0828707
00:55:16 T:12396  NOTICE: hash --> 1446508045.0828707
00:55:16 T:12396  NOTICE: modtime --> 1446508045.1058862
00:55:16 T:12396  NOTICE: hash --> 1446508045.1058862
00:55:16 T:12396  NOTICE: modtime --> 1446508045.084872
00:55:16 T:12396  NOTICE: hash --> 1446508045.084872
00:55:16 T:12396  NOTICE: modtime --> 1446508045.0878751
00:55:16 T:12396  NOTICE: hash --> 1446508045.0878751
00:55:16 T:12396  NOTICE: modtime --> 1446508045.1088893
00:55:16 T:12396  NOTICE: hash --> 1446508045.1088893
00:55:16 T:12396  NOTICE: modtime --> 1446508045.093879
00:55:16 T:12396  NOTICE: hash --> 1446508045.093879
00:55:16 T:12396  NOTICE: modtime --> 1446508045.0658593
00:55:16 T:12396  NOTICE: hash --> 1446508045.0658593
00:55:16 T:12396  NOTICE: modtime --> 1446508045.1088893
00:55:16 T:12396  NOTICE: hash --> 1446508045.1088893
00:55:16 T:12396  NOTICE: modtime --> 1446508045.1008835
00:55:16 T:12396  NOTICE: hash --> 1446508045.1008835
00:55:16 T:12396  NOTICE: modtime --> 1446508045.0988815
00:55:16 T:12396  NOTICE: hash --> 1446508045.0988815
00:55:16 T:12396  NOTICE: modtime --> 1446508045.0898776
00:55:16 T:12396  NOTICE: hash --> 1446508045.0898776
00:55:16 T:12396  NOTICE: modtime --> 1454198114.3555107
00:55:16 T:12396  NOTICE: hash --> 1454198114.284429

marcelveldt:
00:55:16 T:12396   DEBUG: script.skinshortcuts: C:\Program Files (x86)\Kodi_Isengard\portable_data\addons\skin.titan\1080i\script-skinshortcuts-includes.xml has been modified
00:55:16 T:12396   DEBUG: script.skinshortcuts: (1454198114.36 > 1454198114.28)

ghost:
And still a build loop with the same error? If that's the case, on windows something is 'touching' the file after Skin Shortcuts has written it, changing the modified time. As I'm not sure what that is, and unless you have any ideas, its probably worth reverting that commit (assuming a single commit can be reverted - if not I'll do a second commit to undo it, rather than loosing all of the updates from the PR) until I can set up a windows test environment and recreate the problem.

marcelveldt:
Maybe a suggestion could be to check within a time range instead of exact match ?
allowing for example a 2 second difference ? Looks like the difference is only a matter of a few ticks/milliseconds

ghost:
It's possible, but I'd rather not get hacky if I don't have to - especially when I might have figured it out. Can you please edit your xmlfunctions.py - line 622 to:-

```
        for path in paths:
            tree.write( path, encoding="UTF-8" )

            # Save the hash of the file we've just written
            f.close()
            DATA._save_hash( path, "1" )
```

So we ensure the file is closed before we get the hash - should account for any cache the OS has, anything they do on saving, and so on.

marcelveldt:
Changed it but unfortunately it didn't work

ghost:
Damn. So the choices are to revert that commit until I can figure out the problem and fix it, or put in your workaround until and unless I can figure and fix.

The workaround would be line 275: if `("%.2f" % modTime) != ("%.2f" % hash[ 1 ]):` (to check to two decimal places). Your choice which we go for :) (though if its the second, can you test that on Windows first ;))

marcelveldt:
workaround is not working either... Guess the OS is locking or caching the file or something ?

ghost:
In all fairness, looking back at your logs, even 1 decimal place would be too much. I'm going to push for reverting until I can figure it out properly - using the hash of the files has worked well for a long time :)

marcelveldt:
ok, great. Too bad it isn't working as expected. On the other hand, glad that I caught the issue soon before it hits any repo ;-)

marcelveldt:
I think you should be able to revert one single commit from within Github or the git console

ghost:
OK, reverted and thanks for testing :)

marcelveldt:
thanks for the revert. All working again now ;-)

PR #138, opened 2016-02-03
Add requires and templateOnly attribs to custom properties

ghost:
In the Aeon MadNox implementation of Skin Shortcuts, the amount of properties now being written to the main menu has become an issue (causing missing artwork on the home page). This implements two changes which can be used by skins to reduce the number of properties being written to the includes file:-

(1) 'requires' element for custom properties - lets another property be required to be present in the menu for this property to be added to it. E.g. custom widget properties can require the Skin Shortcuts-added 'widget' property

(2) 'templateonly' element for custom properties - with skins using templates, many properties are only useful within the templates. This lets these properties be specified, and then not written to the includes file.

@marcelveldt as this changes how menus are built, how fallbacks are added, and completely how additional properties are handled in the management dialog, if its possible to get this out for testing it would be appreciated. However, as only MadNox will be using it for the foreseeable (and my understanding is MadNox will have its own testing period when the Skin Shortcuts integration is ready), it's not essential if it causes you trouble :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ script.skinshortcuts was written with the intention of making user customizable 
 - Ability to specify an icon to be displayed with a background - [Read More](./resources/docs/advanced/Managing backgrounds.md#defining-backgrounds)
 - Support building variables in templates - [Read More](./resources/docs/advanced/Templates.md#other-template)
 - Ability to specify conditions for fallback custom properties - [Read More](./resources/docs/advanced/Custom shortcut properties.md#for-all-shortcuts-without-a-user-set-property)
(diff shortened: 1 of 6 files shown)


marcelveldt:
I will put this on the beta repo tonight, will allow us for some real life testing.

marcelveldt:
OK, it's now on the beta repo.
Can we have a feature freeze for a couple of days (bugfix only) so we can get a new stable version out early next week ?

ghost:
:+1: for a feature freeze, and thanks for the merge + beta :)

marcelveldt:
One issue discovered on the forums.. Set thumbnail is no longer working for submenus

ghost:
Do you know if thats gui 306 or 311?

ghost:
OK, spotted what's likely causing it to 'no longer work' - I'll provide a PR shortly.

PR #140, opened 2016-02-05
Fix loading skin-provided thumbnail options

ghost:
Possibly fixes issue reported by proxy selecting thumbnails. Two related issues - old code got migrated to the new background thumbnail loading, meaning they never loaded. The code I put in place to stop an infinite loop in case of this happening (as predicted by @marcelveldt :+1:) was flawed, and didn't ensure that an empty list of thumbnails was provided in case the loading of thumbnails failed, causing thumbnail select dialog not to show with gui 311.

@marcelveldt - it's another fairly minor fix, but do you want to get confirmation the fix works before merging?

Edit:- Added a typo fix, which stopped conditional fallbacks correctly matching in the management dialog. Still, at least these bugs are being found before they reach the repo!


marcelveldt:
all working fine now. Thanks!

PR #147, opened 2016-02-16
Minor changes

ghost:
- Krypton changes: Handle new button 7 in DialogSelect.xml; new available shortcuts for TVTimerRules and RadioTimerRules windows
- Context add-on change: Unquote filenames passed from context (change in context addon v0.0.2); unicode fixes for path, label and icon

If we have to make changes for future Krypton support, makes sense to make all current Krypton changes.

### DIFF
--- a/default.py
+++ b/default.py
@@ -254,7 +254,7 @@ class Main:
         self.DEFAULTGROUP = params.get( "defaultGroup", None )
 
         # Properties from context menu addon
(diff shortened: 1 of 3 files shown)


marcelveldt:
Do you want me to review the code ?

ghost:
Whilst a code review from yourself is always welcome, this code should all be safe and so is a case of me forgetting to hit the merge button :blush: Apologies.

PR #149, opened 2016-02-18
Revert "Krypton hotfix"

ghost:
This partially reverts commit a137a6dedb556e2cb6d6402b488c7bd150486e02 - specifically the adding of the "definition" attrib to the built includes.

@phil65 for confirmation that these changes will not be necessary for future Krypton support :)

PR #151, opened 2016-02-20
Build empty templates if not matched

### LINKED ISSUE #142: Low prio: Write empty includes
http://forum.kodi.tv/showthread.php?tid=259210&pid=2239242#pid2239242

When using templates, the XML code is built when needed (the shortcut has condition X or Y).
This works fantastically well for writing widgets to the homemenu and even multiple widgets.

However if the condition is not true for a shortcut, the include is not written at all in the skin, causing some missing include errors in the logs.

For example, in Titan each shortcut can have up to 4 widgets and an include is set for each templated include. Well, when the user only assigned one widget a warning is written to the log.

This error could be prevented ny writing an empty xml element to skin for the include.

Definitely low prio as it's only a warning in the log and does no harm.
I'll look into it myself on a rainy day so this is just a reminder.

ghost:
As discussed in #142 this writes empty includes for any template that isn't built by matching to menu items, avoiding any Skin Has Invalid Includes errors in log.

@marcelveldt hope you don't mind that I went ahead and wrote this, but there are more skins switching to templates so it would be nice to get this feature onto the repo sooner than later. Are you happy with the way I've implemented it?

Closes #142

### DIFF
--- a/resources/lib/template.py
+++ b/resources/lib/template.py
@@ -25,15 +25,20 @@ class Template():
     def __init__( self ):
         # Load the skins template.xml file
         templatepath = os.path.join( __skinpath__ , "template.xml" )
-        self.hasOtherTemplates = False
+        self.otherTemplates = []
         try:
             self.tree = xmltree.parse( templatepath )
 
             log( "Loaded template.xml file")
 
-            # Determine if there are any 'other' templates (used to calculate progress)
-            if len( self.tree.getroot().findall( "other" ) ) != 0:
-                self.hasOtherTemplates = True
+            # Pull out the names and includes of the 'other' templates - used to generate accurate progress
+            # and to build empty 'other' templates if necessary
+            for otherTemplate in self.tree.getroot().findall( "other" ):
+                includeName = "skinshortcuts-template"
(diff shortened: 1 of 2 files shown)


marcelveldt:
haha, you've beaten me to it. Was going to take a look at it one of these days ;-)
Code looking fine, no problems merging this in. Thanks!

ghost:
Cheers. How do you feel, when we know whether #149 is going in, doing a repo push so that the skins adopting templates won't have problems with the mailing list (and so the mailing list find it easier, of course)?

Only one new feature, the rest bug fixes so if a beta period is needed it shouldn't need to be a long one!

marcelveldt:
A repo push sounds good to me. Want to do it now or first revert the temp krypton code ?

ghost:
My reading right of Phil's PR means that the Krypton change won't be necessary, so I say get the reversion in and - if necessary - it can be un-reverted later.

I'd also say get getting-rid of the music library visibility conditions in too. Are you happy doing that?

marcelveldt:
OK, sounds great. I will fix the music visibility then and afterwads we can do another repo pull

PR #155, opened 2016-02-23
remove default Library.HasContent visibility checks for main music and musicvideos node

marcelveldt:
@Ignoble61 as discussed, this removes the visibility conditions for the music shortcuts as it causes a lot of confusion these days


ghost:
Looks good :+1:

PR #159, opened 2016-03-04
fix for the files node in both music and video windows

marcelveldt:
@Ignoble61 it turns out that the shortcut returned by the sscript for the files entry was wrong, this little fix should cover that


ghost:
Can I ask what's wrong with it? I've only tested for videos as my music database needs rebuilding, but [Kodi uses source://video/](http://i.imgur.com/1T2Ieur.png) when navigating from root node to files node, and the shortcut I have that Skin Shortcuts created - ActivateWindow(Videos,sources://video,return) - works...?

marcelveldt:
Yep it works, but it's missing the "add files" entry at the bottom

ghost:
So it is :) What an odd little bug. From a very quick test, it looks like the fact that Skin Shortcuts doesn't append a slash stops that entry from showing up :confused: 

Code looks fine if you want to merge this.

marcelveldt:
yeah, I know. Sounds to me like a Kodfi bug.
Want to merge it in like this or should we just append the trailing slash at all times ?

ghost:
Depends how much time you have :) (I'm just on my lunch break right now!) My personal preference would be to append the trailing slash so that we're consistently using library nodes, rather than mixing and matching paths. But I don't know off-hand whether all node types have or need the trailing slash - that would need testing before merging :)

Or we can merge the fix in now and switch to the slash as and when it's remembered to do so in the future :)

marcelveldt:
You're the (grand)father of the script so your call. I can update the code to make sure there is a trailing slash, that shouldn't harm at all.

ghost:
I would prefer the trailing slash, but only if you have the time to add + test it. More important is working code :)

ghost:
Have you decided what you want to do with this - merge in now, or add the trailing slash?

marcelveldt:
I will go with the slash route, I have time tomorrow night to fix and test it.

ghost:
@marcelveldt It's been about a month since this was last touched. Are you still planning to go the 'slash route' when time allows, or would you prefer it be merged as it is, and the 'slash route' taken in the future?

(If at all possible, I'd like to see an update to the script on the repo in the not too distant future as I and others are increasingly using features only on git, but this fix - or a version of it - really needs to go in before that happens, and for preference time for a beta :))

marcelveldt:
Sorry I did the tests in the meanwhile and all was fine but I still have to write the code and submit a new PR. I was very busy last weekd with work so now I have to catch up with a lot of reports/requests on the forums ;-)

I will see if I get some time tomorrow to put the trailing fix in and prep a new beta on the repo.

Thanks for the reminder!

marcelveldt:
I'll close this one as my local fork has some issues I can't get fixed.
Just reforked and will do another PR

PR #162, opened 2016-03-05
Template improvements

ghost:
A variety of changes to templates. Intended not only to extend functionality, but also allow for much cleaner templates.
- Match multiple elements to a single value - rather than having multiple `<property />` elements which all return the same value, one property element can now match against multiple menu item elements.
- Shared propertyGroup's - When multiple templates use the same `<property />` elements, these can now be pulled out into a single `<propertyGroup />` which is used by multiple templates.
- submenuOther template - Allows building a new 'submenuOther' template based on the properties of a submenu item
- Evaluate mathematical expressions - using [simpleeval](https://github.com/danthedeckie/simpleeval), basic (and not so basic) mathematical expressions can now be evaluated whilst the template is built in a safe manner.

And because I can never resist, one unrelated fix - now actually check if the .hash file exists before trying to parse it. Cleans up log error when the file doesn't exist, whilst leaving relevant logging if there is an issue with the file.

@marcelveldt do you have time to give these changes a quick look over? If not, no worries :) Either way, once they and the other open PR's are merged - and unless you have plans yourself for significant changes for the next release - would you be open to a beta test?

Pinging @mistervee, as I know these changes interest him.

### DIFF
--- a/resources/lib/template.py
+++ b/resources/lib/template.py
@@ -496,52 +496,92 @@ class Template():
                 # Name attrib required, or we've already got a property with this name
                 continue
             name = property.attrib.get( "name" )
-            if "tag" in property.attrib:
-                tag = property.attrib.get( "tag" )
-            else:
-                # No tag property, so this will always match (so let's just use it!)
-                if property.text:
-                    properties[ name ] = property.text
+
+            # Pull out the tag, attribute and value attribs into an array of tuples
+            rules = []
+            propertyValue = ""
+            if "propertyValue" in property.attrib:
+                propertyValue = property.attrib.get( "propertyValue" )
+            
+            # Check for multiple items to match against this single value
+            for singleMatch in property.findall( "rule" ):
+                tag = None
+                attribute = None
+                value = None
+                if "tag" in singleMatch.attrib:
+                    tag = singleMatch.attrib.get( "tag" )
                 else:
-                    properties[ name ] = ""
-                continue
+                    # Tag is required, so we'll pass on this
+                    log( "Trying to match a property without using a tag element" )
+                    continue
+                if "attribute" in singleMatch.attrib:
+                    attribute = singleMatch.attrib.get( "attribute" ).split( "|" )
+                if "value" in singleMatch.attrib:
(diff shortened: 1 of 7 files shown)


marcelveldt:
I've looked at the code which seems to be fine to me, altough I do not really understand the use case (but I'd have to do some testing to better understand). My suggestion is to merge it in and build a new beta version for some actual user testing.

One thing that wasn't really clear to me while quick reading the code: I saw you;ve included the 3th party simple_eval module but it doesn't seem to be used anywhere ?

resources/lib/template.py
```diff
@@ -625,6 +714,32 @@ def replaceElements( self, tree, visibilityCondition, profileVisibility, items,
...
+            if elem.text is not None:
+                while "$MATHS[" in elem.text:
+                    # Split the string into its composite parts
+                    stringStart = elem.text.split( "$MATHS[", 1 )
+                    stringEnd = stringStart[ 1 ].split( "]", 1 )
+                    # stringStart[ 0 ] = Any code before the $MATHS property
+                    # StringEnd[ 0 ] = The maths to be performed
+                    # stringEnd[ 1 ] = Any code after the $MATHS property
+
+                    stringEnd[ 0 ] = simple_eval( stringEnd[ 0 ] )
```
ghost, line 728: simple_eval used here and line 740 ;)

ghost:
I presume you mean use case for `$MATHS[]`...?

I'm using it to automatically position some labels, one for each 'Other' template:

`<top>$MATHS[100 + (50 * $SKINSHORTCUTS[id])]</top>`

so that the labels 'top' position increases by 50 for each menu item. I was also hoping to use it for working around the d*mn disappearing-widgets-if-you-close-the-home-screen-whilst-one-is-loading issue, but the fact that `$MATHS[]` can't do `try: except:` meant that it didn't work in all cases.

I'll resolve the conflicts this evening.

mistervee:
@Ignoble61  Awesome Work. These updates will simplify the widget code a lot. Not sure when I will get to the updates, but I will look at it soonish and provide some feedback.

PR #163, opened 2016-03-09
Simpleeval integration update

ghost:
A couple of issues with the simpleeval integration discovered very quickly :) This PR makes the following changes:-
- Change usage from $MATHS[] to $PYTHON[], to better represent the capabilities of the integration
- Pass in all $SKINSHORTCUTS[] properties so they can be accessed directly by name - this allows properties with square brackets in them to be passed in
- Enable additional operator: 'in'
- Update docs with better examples, list of 'gotcha's'


ghost:
Confirmed working as requested.

PR #165, opened 2016-03-11
Added Android app sources to overrides.xml

sualfred:
This should add Android app sources to the groupings. Sadly untested, because I do not have a Android device right here.


ghost:
Thanks

PR #168, opened 2016-03-17
Add Custom Item option to widgets

ghost:
Adds a 'Custom Item' option to widgets, the same as we have for normal shortcuts, allowing the user to edit the widgetPath property.

The idea is a short-term solution to plugin widgets not refreshing after media plays. With a bit of information, users can now append a Skin Helper reload parameter to a given widget. I'm nervous about adding these automatically as we already know Extended Info needs special consideration, and there may be others - if this is done it needs time to test that it doesn't break widgets in any situation.

@marcelveldt how do you feel about this as an interim solution?

Edit: Updated the PR so the custom item option only shows if the widgetPath property is already set. This should ensure that the additional widget properties are always set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ script.skinshortcuts was written with the intention of making user customizable 
  - Property groups shared between templates - [Read More](./resources/docs/advanced/Templates.md#property-groups)
  - New submenuOther templates - [Read More](./resources/docs/advanced/Templates.md#types-of-templates)
  - Get value from Python - [Read More](./resources/docs/advanced/Templates.md#get-value-from-python)
+- Plugin-based widgets selected via the GUI 312 method will now have a reload parameter added to them automatically. This means that they will update after media has been played. (Requires Skin Helper Service)
 
 #### Version 1.0.7 - repo version
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -6,6 +6,7 @@ v1.0.8
 - New 'visibleoverride'
 - New 'noGroups' build option
 - Template improvements
+- Automatically add a reload parameter to GUI 312 selected widgets
 - Minor fixes
 
 v1.0.7
--- a/default.py
+++ b/default.py
@@ -135,17 +135,31 @@ class Main:
 
             # Load library shortcuts in thread
             thread.start_new_thread( LIBRARY.loadAllLibrary, () )
+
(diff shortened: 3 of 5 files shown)


sualfred:
Will this also be added to the Just Select method? This would save a lot of time writing dialogs for editing the strings.

ghost:
Having thought through this method as-is, it will remove the current additional widget properties. Need to rework it to pass in the current listitem and then return the current properties when the path is edited which I'll do early next week.

@sualfred With the 'Just Select' method you can already add the custom option. With the 'Just Select Widget' method, I'll look into what's needed to add the custom option without messing up the other widget properties when I update this PR.

sualfred:
@Ignoble61 
Great :+1:  Please ping me for testing

ghost:
Improved behaviour so that existing widget properties will not be affected.

Extended to work with Just Select Widget method - when the widget path property is given a skin string to be saved in and that skin string is not empty. Pinging @sualfred to test ;)

ghost:
@marcelveldt - as we've briefly discussed, I've added another update which automatically adds a Skin Helper reload parameter to plugin-based widgets when using GUI 312. In theory, this should solve the (fairly high at the moment) number of reports of widgets not automatically updating. Are you in a position to check the code to see that I'm using the relevant reload parameter correctly?

Also, is everyone happy with the fact that this may well break some plugin based widgets? I've obviously already accounted for Extended Info within the code (possibly not correctly, but I know that needs to be tested fully before any repo push) and will fix any others that are reported to break if they're on the official repo. I'm not going to do any special workarounds for any unofficial plugins, though - but if it does end up breaking any, it could well increase the support-load for both us and skinners utilising the script, even if that support load is just telling people "Well, that's too bad"...

sualfred:
Missed the ping. Sorry :)
Custom entries for Just Select works fine.

ghost:
No need to be sorry :) Thanks for testing. If no objections, and testing with YouTube and other known-troublesome widget providers doesn't throw up any issues, I'll look at merging this in 24 hours or so.

marcelveldt:
Code looks fine to me, I think it's ready to be merged (and tested in beta)

ghost:
Cheers Marcelveldt. I found one typo in last-minute testing which is now fixed. Otherwise all widget providers I tried (which isn't an exhaustive amount) worked as expected, so let's get it merged.

Beta sounds good, though it would be nice to include #159 as part of that ;)

PR #169, opened 2016-03-28
Extend property based on multiple elements to require all rules

ghost:
As requested, allows matching property based on multiple elements to require all the rules specified to match, rather than any.

See new `<match />` element in example below.

```
<property name="leftartwork" propertyValue="5">
    <rule tag="property" attribute="name|widgetStyle" value="Panel" />
    <rule tag="property" attribute="name|widgetArt" value="Poster" />
    <rule tag="property" attribute="name|widgetCase" value="Glass" />
    <match>all</match
</property>
```

Pinging @mistervee to test :)

### DIFF
--- a/resources/docs/advanced/Templates.md
+++ b/resources/docs/advanced/Templates.md
@@ -202,6 +202,10 @@ You can set a property to a single value based on different matching elements wi
 
 Would set the property `$SKINSHORTCUTS[leftartwork]`to the value of `5` if any of the three rules match.
 
+By default, this property will match if any of the rules match. To only match the property if all of the rules match, include `<match>all</match>` as a sub element.
+
+Note, when setting a property based on multiple elements, you /must/ specify the propertyValue you want assigned - you cannot assign the property to the value of the menu items element that you are checking against.
+
 #### Fallback value
 
 If none of the `<property />` elements match, the property will be an empty string. You can set an alternative fallback value by including the following as the final element in a list of `<property />`'s
--- a/resources/lib/template.py
+++ b/resources/lib/template.py
@@ -537,14 +537,17 @@ class Template():
     def getProperties( self, elem, items ):
         # Get any properties specified in an 'other' template
         properties = {}
+
+        # Start by finding all properties defined directly in the template
         searchProperties = elem.findall( "property" )
 
-        # Start by finding any global property groups
+        # Add any properties defined in a property group
         for propertyGroup in elem.findall( "propertyGroup" ):
             for searchGroup in self.tree.findall( "propertyGroup" ):
                 if propertyGroup.text.lower() == searchGroup.attrib.get( "name" ).lower():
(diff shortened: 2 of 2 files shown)


mistervee:
@Ignoble61 Thanks. Working well.

The one thing I did notice is if I do this and the first multiple property element is matched the second value is used. I actually expected the first property match to be used and the script to ignore the second.

`<property name="leftart" propertyValue="4.5">
`
`<rule tag="property" attribute="name|widgetArt" value="Poster" />
`
`<rule tag="property" attribute="name|widgetCase" value="Case" />
`
`<match>all</match>
`
`</property>
`
`<property name="leftart">2.5</property>`

Not a big deal for me, I can work with it as you have designed.

ghost:
It should indeed only match the first one. I'll try and track down that bug tonight, otherwise I'll push a fix tomorrow evening.

ghost:
@mistervee I've reviewed the code and can't just see where it's going wrong, so can you please try again with the latest commit (which adds extra logging) and post a link to a log - with Skin Shortcuts own debug logging option enabled. Hopefully that will at the very least track down where the issue is. Cheers.

mistervee:
http://pastebin.com/7La1j7Pv
here is the log.

The logging addition has seemed to of broken the template.

Here is the madnox template xml I am working on
https://onedrive.live.com/redir?resid=AE1918D112D1916C!5362&authkey=!AB3ZDT1iAp86yVA&ithint=file%2cxml

mistervee:
The code in question is around line 500.

Only widget 1, Panel Style, Poster and square has the new code.

ghost:
Hopefully spotted the issue - I didn't take into account that the script has to check each element of the menu item individually to see if it has the same property as the rule wants to match against, so the code was thinking that the rules weren't matching far too soon.

If the fix doesn't work, another debug log please :)

By the way - and assuming I've done my job correctly, and its entirely possible I haven't - you can access your properties directly in Python snippets by name, so

`$PYTHON[((300 - int($SKINSHORTCUTS[containerheight])) / 2) + 454]`

should be able to be just

`$PYTHON[((300 - int(containerheight)) / 2) + 454]`

mistervee:
http://pastebin.com/wp4nS1Ee

The script seems to stall on building the properties. It only matches the 1st one and then no other properties build after.

ghost:
OK, one more attempt now uploaded, and when it still doesn't work I'll look properly into it tomorrow :)

(Thanks for the thorough testing, by the way!)

mistervee:
Working great now.

Sadly this does not work without the $SKINSHORTCUTS
`$PYTHON[((300 - int(containerheight)) / 2) + 454]`

No trouble at all testing this out. A big thanks for incorporating the feature.

ghost:
I'll add that to my list of things to debug for another day ;)

PR #171, opened 2016-03-30
Update PlayDVD action

ghost:
Based on [this thread](http://forum.kodi.tv/showthread.php?tid=266709).

Looks like either the action for playing a dvd has been wrong since the script was created, or the action has changed at some point. Does anyone reading this have a DVD player in their system and can test this?


ghost:
As no-one has been able to test, I'm going to go with what the wiki says and merge this :) It's either been broken for a while, or will be picked up during a beta.

PR #176, opened 2016-05-03
Update Hebrew Translation

tomer953:
it seems that the language team has abandoned this add-on (no update for
more than a year)  so I translated It myself.


ghost:
Thanks for this.

PR #179, opened 2016-05-15
Updates

### LINKED ISSUE #127: request: option to hide shortcut
http://forum.kodi.tv/showthread.php?tid=205482&pid=2216301#pid2216301

ghost:
With 1.0.8 having been a relatively large update, hopefully 1.0.9 will be a much simpler release :) Initial updates:-
- Remove `__variable__` style naming (also removes a number of 'static' vars that are not used)
- Introduce context menu for management dialogue
- Introduce GUI 313 to disable/enable menu items, and add it by default to context menu for GUI 211 (closes #127)

It's important to note that GUI 211 isn't used directly by a number of skins (but rather 303/304 are focused), and so the context menu and GUI 313 won't be accessible without modifications on these skins which has the potential to cause support issues. @marcelveldt I said I might have to pick your brains on this - if you have any suggestions for how this should be handled, please say :)

In line with keeping this a minor update, the only further changes I plan are requests (currently to add default icons to all available shortcuts so they may be overriden) and Krypton changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ script.skinshortcuts was written with the intention of making user customizable 
 #### Version 1.0.9 - git version
 
 - Ability to move controls from management dialog to context menu (Krypton only) - [Read More](./resources/docs/advanced/Context menu.md)
+- Ability to disable menu items via GUI 313 - Please note this control is made available by default on the context menu - [Read More](./resources/docs/started/Management Dialog.md)
 
 #### Version 1.0.8 - repo version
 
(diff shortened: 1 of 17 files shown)


ghost:
Default icons added, and Krypton changes made.

marcelveldt:
@Ignoble61 do you want me to test something in particular ?

ghost:
Mike Silvo and Nox 5 Silvo users (who have been vocal in wanting the disable button) were very helpful in testing the fundamentals of the context menu and GUI 313, so they don't need testing per se, but if you have any feedback or ideas on the default settings for the context menu (on GUI 211, contains GUI 313) they would be most welcome - I believe I'm right in saying the context menu won't appear by default in Titan (just as it doesn't in igConq), but I couldn't come up with better defaults whereby it would, but wouldn't show up on inappropriate controls on other skins...

PR #185, opened 2016-06-25
Krypton window fixes

ghost:
Krypton has removed several window ID's that were previously deprecated that this script still accesses. This PR has two functions - to update library functions so that they return they correct window ID's after these changes (Videolibrary > Videos, MusicLibary > Music) and to update any existing shortcuts to use these window IDs (most notable in this case, widgets whose widgetTarget is 'video' get upgraded to 'videos').

For full disclosure, due to real life commitments this has largely been coded blind meaning it has had very limited testing in any real life situation. However with people beginning to notice the script is broken on the latest nightlies I've made the decision to create the PR ahead of being able to test it thoroughly myself.

@marcelveldt are you in a position to test whether this solves the window issues created by recent Krypton changes? If so, is it possible to get an updated beta on the Emby beta repo?

If not, I will test these changes thoroughly early next week and merge if everything works as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,12 +5,13 @@ script.skinshortcuts was written with the intention of making user customizable 
 
 ## What's New for Skinners
 
-#### Version 1.0.9 - git version
+#### Version 1.0.10 - git version
 
 - Ability to move controls from management dialog to context menu (Krypton only) - [Read More](./resources/docs/advanced/Context menu.md)
 - Ability to disable menu items via GUI 313 - Please note this control is made available by default on the context menu - [Read More](./resources/docs/started/Management Dialog.md)
 - Default icons for all available shortcuts in the Common, Commands and Settings available shortcut groupings - [Read More](./resources/docs/advanced/Overriding icons.md)
 - Behaviour change: When using mode=single, only the main menu (and corresponding submenu's) will be added to the allmenus include - additional groups that are being built will no longer be added
+- Please be aware of the removal of deprecated window names in Krypton. The script will attempt to update VideoLibrary and MusicLibrary to Videos and Music respectively, but you should ensure that any default shortcuts you provide with your skin are up to date as this behaviour may change in the future.
 
 #### Version 1.0.8 - repo version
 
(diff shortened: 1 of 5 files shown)


MikeSiLVO:
@Ignoble61 I can test anything you need. Just let me know what, if anything, you would like tested :smiley:

sualfred:
Just gave it a quick try. Everything seems to be okay and working as far as I can tell.

ghost:
@mikesilvo164 - thanks for the offer. Shortcuts to everything video-library wise needs testing (playlists, nodes, addons, etc) as well as video library widgets (that is to say widgets which open the video library when clicked on, not that start to play a video) - everything else should work with the current version but, obviously, if there is anything broken...

@sualfred - thanks for testing :)

marcelveldt:
@Ignoble61 based on the initial test reports here and on the forum thread I'd say merge this and I'll put a beta version on the emby repo to have some real life testing.

Sorry for the late reply btw, as I told you before at summertime I don't spend too much time behind the pc ;-)  

I will do some testing on my own too once it's merged.

ghost:
I'd agree - It's had some good community testing and I've just run most of the tests I wanted to do, however backwards compatibility (Isengard/music is a potential concern) still needs to be checked but I've run out of time to test this now. I'll quickly merge the commits and then merge it :)

No need to be sorry at all, enjoy the summer!

MikeSiLVO:
@Ignoble61 I just had a chance to do some testing and the only thing I noticed is that MusicLibrary links in the submenu of MadNox did not change to Music but they still works so not sure if it is relevant.

ghost:
Mike, the underlying data.xml files aren't updated so if a shortcut is to MusicLibrary (or VideoLibrary), they'll still show MusicLibrary/VideoLibrary in the menu editor - when the menu is built, though, the final includes.xml should have the correct window used.

MikeSiLVO:
Ah, I see... Wasn't sure if it was important since that is superficial and all, however I thought the extra info couldn't hurt.

ghost:
All feedback is very very welcome :) (And if I find a good way of doing it, I'll fix the superficial stuff as well!)

MikeSiLVO:
Not sure if you even need to waste time on that. If the include file gets updated and any new shortcuts added use the proper window names going forward then as long as the default links in the skin are updated then it would probably only "bug" me that the names in the dialog aren't right. I already changed them manually before the script update :wink:

Side note: I hope the Krypton skinning changes slow down now... Getting to the point one could feel it might be easier to start a skin over from scratch. Looking forward to enjoying the media instead of updating the skin, at least for a little while, but without my hobby I'd probably get really bored :smiley:

ghost:
First Krypton beta is due in a matter of days (first of next month?), so they really should be all but done with the skinning engine changes :)

PR #199, opened 2016-10-22
Use Krypton inbuilt select dialog

ghost:
Updates the select dialog calls which display an icon along with the item (internally referred to as the 'pretty' select dialog) to use the new useDetails=True call introduced with Krypton.

@marcelveldt I have tested this locally, but I'm still restoring my system after an epic juice-related hardware failure so I wouldn't class my tests as definitive. Are you in a position to test + review this before it's merged?


marcelveldt:
image

marcelveldt:
Don't know if you have access to my codacy details but there were spotted a few small issues with the PR... See screenshot

ghost:
Liking the codacy :)

Unfortunately it's highlighting something I missed entirely whilst coding this, which is that we can't (seem) to use the 'Get More...' button in the select dialog with the new method. I'm going to close this for now until I think about how to deal with this.

Cheers

phil65:
I would stay with your old method until we implement a proper way to deal with the "get more..." button.
I can take a look at that (shouldnt be that hard to implement), but that´s for v18 then..

phil65:
Btw most codacy errors should also get recognized by some of the common Python linters out there. ;)

ghost:
And when I've got this system fully re-installed (so I'm not just coding in notepad whilst I await the replacement hdd) I'll get back to having proper debug tools :) Cheers Phil, anything you could do for the Get More would be great.

PR #201, opened 2016-11-05
Template and additional submenu improvements

ghost:
Submenu templates:-
* Add 'sub templates' which build for each item in the submenu and can then be inserted into the primary submenu template which is built just once for each submenu.

Additional submenus:-
* Add ability for skin to indicate that the additional submenu levels shouldn't be shared between skins (this allows me to use them in a way that won't be compatible with other skins)
* Various fixes related to additional submenu's, most notably corrects a bug where the .DATA.xml and the associated properties wouldn't be copied from skin defaults when first editing the menu

@marcelveldt I've included a 'Mark stable' commit in this as I'd like to get this onto the repo sooner than later as I'm already using these features. Would you prefer to go through a public test first? If so, I'll remove that commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,17 @@
-#Skin Shortcuts - 1.0.11
+#Skin Shortcuts - 1.0.12
 
 script.skinshortcuts was written with the intention of making user customizable shortcuts on the home page easier for skinners.
 
 
 ## What's New for Skinners
 
-#### Version 1.0.11 - git version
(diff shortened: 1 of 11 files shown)


ghost:
As there seems to be no objection I'm going to merge this now and look at a repo push later today. If there are any issues with this, there's normally at least a couple of days to cancel the repo update request.

PR #205, opened 2016-11-22
added some support for resource addons

marcelveldt:


addon.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.skinshortcuts" name="Skin Shortcuts" version="1.0.12" provider-name="BigNoid,marcelveldt">
+<addon id="script.skinshortcuts" name="Skin Shortcuts" version="1.0.13" provider-name="BigNoid,marcelveldt">
```
ghost: Can you please bump to .14, as .13 will be for the fix for the bug I introduced in .12 and will need to be requested for the repo sooner than later :)

resources/lib/gui.py
```diff
@@ -1129,7 +1131,7 @@ def onClick(self, controlID):
             # Display list items
             self._display_listitems( focus = itemIndex )
         
-        elif controlID == 301:
+        elif controlID in [301, 1301]:
```
ghost, line 1153: Does gui 1301 need to be documented?

ghost:
Am I right in thinking this means that if the thumbnail image selected is from an image resource pack, it will be saved in the .DATA.xml file, and so shared between skins? If so, very nice :)

(And it looks good even if that's not the case ;))

marcelveldt:
Yeah, that's right. I finished my major refactor of skinhelper and long the road added some new stuff
New is for example a new image picker which also lets you pick from resource addons.

that image picker can also be used with skinshortcuts now

marcelveldt:
OK, I'll bump the version tomorrow.

resources/lib/gui.py
marcelveldt: Yeah I think it does. I will take some time for that next week if you don't mind. Currently busy replying to all (bug) reports on the forums after my big code rewrite of skinhelper ;-)

marcelveldt:
We'll merge this one after .13 hits the repo.

ghost:
Assuming I hear nothing negative about the update, I'll be requesting .13 in the morning. (The new image picker sounds great!)

ghost:
@marcelveldt Could you please update the labels for Smart Shortcuts and Default Widgets in the overrides.xml - these both rely on strings from Skin Helper which are removed in your new version. Thanks.

marcelveldt:
Oh yes that's right, just fixed it.

ghost:
Thanks for updating the strings. 1.0.13 is now on the repo, so this can be merged whenever you're ready.

marcelveldt:
Let's wait a few more days while users are testing my huge update on my (new!) own beta repo. If there are no bug reports regarding skinshortcuts we can merge this into main.

BTW: If you have some time available take a look at the latest beta of titan to see the new image picker in action with the resource addon support. I think it's pretty cool.

ghost:
I've had a play and it's a great feature :)

One question about the new Skin Helper Service - I notice the widgets are now a separate script (script.skin.helper.widgets). From what I can see installing the primary service doesn't automatically install this module, so how would you recommend handling 'Enable default widgets'/'Get More...' within Skin Shortcuts current widget select?

('Enable default widgets' installs Skin Helper so that it can generate the list, add Skin Helper Widgets to the 'Get More...'?)

marcelveldt:
Option 1) Set skinhelper service as requirement for skinhelper widgets.

Option 2) Move the widget properties code and skinshortcuts code to skinhelper widgets

Option 3) Move widget properties to skinhelper widgets addon and move the skinshortcuts methods to skinshortcuts script itself

IMO option 3 is the most clean version. I already stripped all code as such that it is just one single file in skinhelper service (skinshortcuts.py) that handles all logic to interact to skinshortcuts. So maybe it's a good plan to just integrate it into skinshortcuts ?

ghost:
I'm always a fan of keeping related code in the same place, so #3 seems the cleanest to me. But this is your code, so I'm very happy to defer to your final judgement :)

(I'm away for a few days from the morning, but if you need any assistance moving code then send me a PM and I'll act on it as soon as I'm able :))

ghost:
Just a general comment which shouldn't affect the code at this point, though it's worth being aware of when assessing feedback of this update - there were two reasons why custom thumbnail images were moved to the .properties file rather than the .DATA.xml file

(1) Some images selected were provided by the skin, and so weren't available when switching to a different skin (obviously doesn't affect the new usage of resource addons)
(2) Different skins have different styles, so some users preferred to have different icons dependant on which skin they were using - worth being aware of when these updates get into wide-usage. Not necessarily an issue but - _if_ it becomes so - possibly worth considering whether to keep all thumbnails, whether resource add-on or not, skin-specific.

Also worth noting that, if I'm reading the code correctly (and I apologise that I don't have time to test this before the weekend), from looking at the code if a thumbnail image is already selected (via the skins .properties file) this will take precedence over any newly selected resource image (as all images in the .properties file do, over the .DATA.xml set images) - this would require calling _remove_additionalproperty() before saving a property directly in the .DATA..xml file.

ghost:
@marcelveldt is it possible for you to rebase this on master? Despite the fact the Kodi 18 is still early-alpha, there are plenty of reports on the forums of people using it and the basic-compatibility in master aren't included in your builds so people using your repo are having to manually install the master branch. Cheer.

marcelveldt:
will do!

PR #209, opened 2016-12-14
Replace IsEmpty with String.IsEmpty on Kodi 17+

ghost:
Should fix disabled controls in management dialog, single-list menu and automatic visibility condition for Weather when using v18.

Ping @BigNoid - should fix the issue you reported on the forum.

### DIFF
--- a/default.py
+++ b/default.py
@@ -38,6 +38,10 @@ DATA = datafunctions.DataFunctions()
 LIBRARY = library.LibraryFunctions()
 NODE = nodefunctions.NodeFunctions()
 
+ISEMPTY = "IsEmpty"
+if int( xbmc.getInfoLabel( "System.BuildVersion" ).split(".")[0] ) >= 17:
+    ISEMPTY = "String.IsEmpty"
+
 hashlist = []
 
 def log(txt):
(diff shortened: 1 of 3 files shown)


BigNoid:
Thx @Ignoble61 this pr fixes the issue mentioned on the forums.

Guilouz:
@Ignoble61 @BigNoid 
In default.py, why not just change Isempty with String. IsEmpty ?

When I try your changes in default.py, my mainmenu and powermenu not loading but they are restore my problem with ID 401, 307 and 402 controls in management dialog.

I try to just replace IsEmpty by String.IsEmpty and all is working fine with latest Kodi 18 build.

ghost:
Because String.IsEmpty was only introduced in Krypton, and the script supports older versions of Kodi than that.

Guilouz:
@Ignoble61 Ok but this changes not working correctly. I have no mainmenu/powermenu on my home screen.

ghost:
Latest version of Estuary Mod is working fine for me with these changes, so please go back to the script thread and post the debug log that was requested if these changes didn't work. Also include details of how to reproduce - any particular settings that the issue occurs with, for example. Help us help you :)

phil65:
Just an idea: Why not sending different script versions to our various version branches? That way you wouldnt have to deal with bw compatibility within your script.

ghost:
This is the first time I can remember where there are changes that break this script, so it's definitely worth considering splitting the v18 release of the script. At this point, though, there's no v18 repo that I can see (though there are a fair few v18 users) - so by fixing the breaking changes in core now, any such users would get those if the script is updated on the repo in the meantime.

There's also missing api features for v18 (files.getsources can't get game sources yet, for example), which mean that whilst a v18 split can be made, a proper fully-supporting-v18 version can't actually be written yet ;)

phil65:
you can do the split for krypton repository already since that one supports "string.isempty" etc.

ghost:
Fair point. From my time-availability, something to be considered in the new year :)

phil65:
sure. In the end, do what best fits your workflow etc.. just wanted to mention it.

ghost:
Your input is always appreciated.

PR #214, opened 2017-01-17
Update strings.po

abeksis:
Missing a few things in your new version, so I added and translated into Hebrew.


marcelveldt:
thanks!

PR #218, opened 2017-02-20
resources/language/Hebrew/strings.po

abeksis:
the file send you need to be strings.po not strings-he.po

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ It's highly recommended to take time to read through these documents before you 
 
 If you require any assistance post on the Kodi forum and I'll do my best to assist:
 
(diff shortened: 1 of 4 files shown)


ghost:
This PR appears to add a new file called string-he.po ...? AFAIK that isn't the correct name for the file.

ghost:
Whilst the additional commit of allowing thumbs to be set as None through property makes sense, I'm still not sure of the benefit of adding /Hebrew/strings-he.po when Kodi should load the relevant strings.po file...?

marcelveldt:
@abeksis please resubmit your PR rebased on latest code as the current PR is not properly merged/squashed. Also the language file is in the wrong location

PR #219, opened 2017-03-03
tiny fix: assets tags need to be inside extension tags

mgonzales71:
just happen to notice this - figured it would be easier to just propose a quick PR :smile: 
cheers!


ghost:
Nice spot, and many thanks :)

PR #222, opened 2017-03-12
Hide home button if on home

BADMS:
I can't see a way to create a custom 'visible' for menu items, so in this case the home button is useless on homepage. IF there is a way to do this then ignore [apart from making another menu just for home?].

### DIFF
--- a/resources/lib/datafunctions.py
+++ b/resources/lib/datafunctions.py
@@ -873,6 +873,8 @@ class DataFunctions():
             return "[System.Platform.Windows | System.Platform.Linux] +! System.Platform.Linux.RaspberryPi"
 
         # General visibilities
+        elif action == "activatewindow(home)":
+            return "!Window.IsVisible(home)"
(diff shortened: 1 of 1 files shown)


ghost:
```
<override action="ActivateWindow(Home)">
    <condition>!Window.IsVisible(Home)</condition>
<override>
```

BADMS:
I tried the code, it has no affect. I still see items on homepage with the overrides set in overrides.xml?

ghost:
Sorry this has taken me so long to look at. The code snippet I provided works for me, so can you please provide a full debug log, an overrides.xml showing the proposed solution in place and a built script-skinshortcuts-includes.xml that shows that you still have the issue with the proposed workaround in place.

Once I've received and reviewed that, I'll happily merge this as a workaround to a bug I can't reproduce :)

ghost:
Is this PR still relevant?

BADMS:
sorry, forgot this. will check when possble thanks

